PR #5, opened 2017-11-17
Turn off implicit-casts and implicit-dynamic

jacehensley-wf:
## Ultimate Problem
This will be how Dart 2 operates, might as well follow those guidelines

## Solution
- Turn off `implicit-casts` and `implicit-dynamic`
- Move code around to better allow for this

## Testing Suggestions
- Verify tests pass

Everything

---
FYA: @evanweible-wf

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -15,15 +15,12 @@
 import 'dart:io';
 
 import 'package:logging/logging.dart';
+import 'package:yaml/yaml.dart';
+
+export 'package:yaml/yaml.dart' show loadYaml;
 
 final RegExp importExportPackageRegex =
     new RegExp(r'''^(import|export)\s+['"]package:([a-zA-Z_]+)\/.+$''', multiLine: true);
-
-const dependenciesKey = 'dependencies';
-const dependencyValidatorPackageName = 'dependency_validator';
-const devDependenciesKey = 'dev_dependencies';
-const nameKey = 'name';
-const transformersKey = 'transformers';
 
 final Logger logger = new Logger('dependency_validator');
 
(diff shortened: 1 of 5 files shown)


lib/src/utils.dart
```diff
@@ -15,28 +15,63 @@
...
+
+  String get name => _yamlMap[_nameKey] as String;
+
+  Set<String> get dependencies =>
+      ((_yamlMap[_dependenciesKey] as YamlMap ?? const <dynamic, dynamic>{}).keys as Iterable<String>).toSet();
+
+  Set<String> get devDependencies =>
+      ((_yamlMap[_devDependenciesKey] as YamlMap ?? const <dynamic, dynamic>{}).keys as Iterable<String>).toSet()
+        // Remove this package, since we know they're using our executable
+        ..remove(_dependencyValidatorPackageName);
```
evanweible-wf: Filtering out the `dependency_validator` package from the dev deps seems like it should happen elsewhere, since this is just a generic `PubspecYaml` class to help parse a `pubspec.yaml` file.

lib/src/utils.dart
```diff
@@ -15,28 +15,63 @@
...
+
+  Set<String> get packagesUsedViaTransformers {
+    final transformerEntries = _yamlMap[_transformersKey] as Iterable<Object>;
```
evanweible-wf: Similarly, this seems to me like it should expose the transformer configurations in a typed manner, and then a separate utility function could take a `PubspecYaml` instance as input and return the package names

lib/src/utils.dart
jacehensley-wf: Hmm, I meant this class to be specific to this use-case. That's why I made it private. I can rename it if you'd like

evanweible-wf: yeah I guess it's private so it's not a big deal

jacehensley-wf:
Will re-open when this is more important / when Dart 2 comes out.

PR #7, opened 2017-11-29
WP-5430 Add special message when depending on the analyzer package but not us…

jacehensley-wf:
## Ultimate Problem
- There were some cases before when packages were declaring `analyzer` as a dependency because they thought it was necessary to run the analyzer cli. This would cause unnecessary dependency locking.

## Solution
- Add a special message calling out that behavior
  - Don't throw, because that would be breaking, also we can't be sure that's what's happening.

## Testing Suggestions
- Verify tests pass

- N/A

---
FYA: @evanweible-wf @maxwellpeterson-wf @jayudey-wf 
FYI: @greglittlefield-wf

### DIFF
--- a/lib/dependency_validator.dart
+++ b/lib/dependency_validator.dart
@@ -173,7 +173,8 @@ void run({List<String> ignoredPackages = const []}) {
             ..remove(dependencyValidatorPackageName);
 
   if (unusedDependencies.contains('analyzer')) {
-    logger.warning('You do not need to depend on the analyzer to run the dart analyzer.');
+    logger.warning('''You do not need to depend on `analyzer` to run the Dart analyzer.
+        Instead, just run the `dartanalyzer` executable that is bundled with the Dart SDK.''');
   }
 
   unusedDependencies
--- a/lib/dependency_validator.dart
+++ b/lib/dependency_validator.dart
@@ -173,8 +173,10 @@ void run({List<String> ignoredPackages = const []}) {
             ..remove(dependencyValidatorPackageName);
 
   if (unusedDependencies.contains('analyzer')) {
-    logger.warning('''You do not need to depend on `analyzer` to run the Dart analyzer.
-        Instead, just run the `dartanalyzer` executable that is bundled with the Dart SDK.''');
+    logger.warning(
+      'You do not need to depend on `analyzer` to run the Dart analyzer.\n'
+          'Instead, just run the `dartanalyzer` executable that is bundled with the Dart SDK.',
+    );
   }
 
   unusedDependencies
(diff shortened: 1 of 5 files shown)


aviary-wf:
## Raven
<details>
<summary>Number of Findings: 0</summary>

</details>

test/executable_test.dart
```diff
@@ -73,6 +73,13 @@ void main() {
       expect(result.stderr, contains('dart_dev'));
     });
 
+    test('warns when the analyzer pacakge is ', () {
```
evanweible-wf: I'm assuming this should be "does not warn when the analyzer dependency is used" and there should be a second additional test that is "warns when the analyzer dependency is unused"?

lib/dependency_validator.dart
```diff
@@ -170,9 +170,15 @@ void run({List<String> ignoredPackages = const []}) {
           // Remove all deps being used for their transformer(s)
           .difference(packagesUsedViaTransformers)
             // Remove this package, since we know they're using our executable
-            ..remove(dependencyValidatorPackageName)
-            // Ignore known unused packages
-            ..removeAll(ignoredPackages);
+            ..remove(dependencyValidatorPackageName);
+
+  if (unusedDependencies.contains('analyzer')) {
+    logger.warning('You do not need to depend on the analyzer to run the dart analyzer.');
```
greglittlefield-wf: #nit wording is a little off in `on the analyzer`, `Dart` is not capitalized

Might also be nice to say what you should do instead.
```
You do not need to depend on `analyzer` to run the Dart analyzer.
Instead, just run the `dartanalyzer` executable that is bundled with the Dart SDK.
```

lib/dependency_validator.dart
```diff
@@ -170,9 +170,16 @@ void run({List<String> ignoredPackages = const []}) {
...
           .difference(packagesUsedViaTransformers)
             // Remove this package, since we know they're using our executable
-            ..remove(dependencyValidatorPackageName)
-            // Ignore known unused packages
-            ..removeAll(ignoredPackages);
+            ..remove(dependencyValidatorPackageName);
+
+  if (unusedDependencies.contains('analyzer')) {
+    logger.warning('''You do not need to depend on `analyzer` to run the Dart analyzer.
+        Instead, just run the `dartanalyzer` executable that is bundled with the Dart SDK.''');
```
evanweible-wf: fyi, this indents this line 8 spaces, was that intended?

```
Validating dependencies for dependency_validator
You do not need to depend on `analyzer` to run the Dart analyzer.
        Instead, just run the `dartanalyzer` executable that is bundled with the Dart SDK.
These packages may be unused, or you may be using executables or assets from these packages:
  * analyzer
  * coverage
  * dart_style
  * dartdoc
```

maxwellpeterson-wf:
+10
- CI passes

maxwellpeterson-wf:
QA +1
- [x] Dev +1's
  - Documented [here](https://github.com/Workiva/dependency_validator/pull/7#issuecomment-354894836)
- [x] Tests created/updated 
- [x] All tests pass

Issue #8, opened 2018-01-19
Add --exclude-dir option to allow excluding certain directories


evanweible-wf:
Some projects have nested packages in `example/` or something similar which can lead to inaccurate results when running dependency_validator. We should allow users to exclude directories to address this issue.

greglittlefield-wf:
Fixed in https://github.com/Workiva/dependency_validator/pull/15 (version 1.1.0)

PR #15, opened 2018-02-23
UIP-3068 Add ability to exclude directories when checking for infractions

jacehensley-wf:
## Ultimate Problem
https://github.com/Workiva/dependency_validator/issues/8

## Solution
- Add the `--exclude-dir` flag

## Testing Suggestions
- Verify tests pass

- Error reporting

---
FYA: @greglittlefield-wf @maxwellpeterson-wf @toddbeckman-wf 
FYI: @evanweible-wf

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -31,18 +31,15 @@ String bulletItems(Iterable<String> items) => items.map((l) => '  * $l').join('\
 
 Iterable<File> listDartFilesIn(String dirPath, List<String> excludedDirs) {
   if (!FileSystemEntity.isDirectorySync(dirPath)) return const [];
-  final files = new Directory(dirPath)
-      .listSync(recursive: true)
-      .where((entity) {
-        if (entity is! File) return false;
-        if (entity.path.contains('/packages/')) return false;
-        if (!entity.path.endsWith('.dart')) return false;
-        if (excludedDirs.any(entity.path.startsWith)) return false;
 
-        return true;
-      });
+  return new Directory(dirPath).listSync(recursive: true).where((entity) {
+    if (entity is! File) return false;
+    if (entity.path.contains('/packages/')) return false;
+    if (!entity.path.endsWith('.dart')) return false;
+    if (excludedDirs.any(entity.path.startsWith)) return false;
 
-  return files;
+    return true;
+  });
 }
 
 void logDependencyInfractions(String infraction, Iterable<String> dependencies) {
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -15,6 +15,7 @@
 import 'dart:io';
 
 import 'package:logging/logging.dart';
+import 'package:path/path.dart' as p;
 
 final RegExp importExportPackageRegex =
     new RegExp(r'''^(import|export)\s+['"]package:([a-zA-Z_]+)\/.+$''', multiLine: true);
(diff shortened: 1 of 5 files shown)


aviary-wf:
## Module Findings
No security relevant content detected. Please review for security relevance and request security review as needed.

lib/src/utils.dart
```diff
@@ -29,11 +29,17 @@ final Logger logger = new Logger('dependency_validator');
...
   if (!FileSystemEntity.isDirectorySync(dirPath)) return const [];
-  return new Directory(dirPath)
-      .listSync(recursive: true)
-      .where((entity) => entity is File && !entity.path.contains('/packages/') && entity.path.endsWith('.dart'));
+
+  return new Directory(dirPath).listSync(recursive: true).where((entity) {
+    if (entity is! File) return false;
+    if (entity.path.contains('/packages/')) return false;
+    if (!entity.path.endsWith('.dart')) return false;
+    if (excludedDirs.any(entity.path.startsWith)) return false;
```
greglittlefield-wf: This could cause false positives if a directory is a substring of the excluded dir and the user forgets a trailing slash.

You can use [`path.isWithin`](https://www.dartdocs.org/documentation/path/1.5.1/path/isWithin.html) to handle this.

lib/src/utils.dart
```diff
@@ -29,11 +29,17 @@ final Logger logger = new Logger('dependency_validator');
...
-Iterable<File> listDartFilesIn(String dirPath) {
+Iterable<File> listDartFilesIn(String dirPath, List<String> excludedDirs) {
   if (!FileSystemEntity.isDirectorySync(dirPath)) return const [];
-  return new Directory(dirPath)
-      .listSync(recursive: true)
-      .where((entity) => entity is File && !entity.path.contains('/packages/') && entity.path.endsWith('.dart'));
+
+  return new Directory(dirPath).listSync(recursive: true).where((entity) {
+    if (entity is! File) return false;
+    if (entity.path.contains('/packages/')) return false;
```
greglittlefield-wf: Path #nit; it would be more portable to do:
```dart
if (path.split(entity.path).contains('packages')) return false;
```

https://www.dartdocs.org/documentation/path/1.5.1/path/split.html

test/executable_test.dart
```diff
@@ -66,6 +68,13 @@ void main() {
         expect(result.stderr, contains('These packages are used in lib/ but are not dependencies:'));
         expect(result.stderr, contains('yaml'));
       });
+
+      test('expect when the lib directory is excluded', () {
```
greglittlefield-wf: Typo: should be `except`

lib/src/utils.dart
jacehensley-wf: whoa... Dart never ceases to amaze me

lib/src/utils.dart
```diff
@@ -34,9 +35,9 @@ Iterable<File> listDartFilesIn(String dirPath, List<String> excludedDirs) {
 
   return new Directory(dirPath).listSync(recursive: true).where((entity) {
     if (entity is! File) return false;
-    if (entity.path.contains('/packages/')) return false;
-    if (!entity.path.endsWith('.dart')) return false;
-    if (excludedDirs.any(entity.path.startsWith)) return false;
+    if (entity.path.split(p.separator).contains('packages')) return false;
```
greglittlefield-wf: This is probably fine but in the future I'd recommend using `p.split`, and not `.split(p.separator)`.

For example:
```dart
// Multiple slashes should collapse to one in most cases
print('foo//bar'.split(p.separator)); // [foo, , bar]
print(p.split('foo//bar'));           // [foo, bar]

// In Windows, you can mix slashes
print(r'foo/bar\baz'.split(p.windows.separator)); // [foo/bar, baz]
print(p.windows.split(r'foo/bar\baz'));           // [foo, bar, baz]
```

Paths have so many edge-cases, and it's usually better to not make any assumptions and let the `path` package take care of it for you.

maxwellpeterson-wf:
+10
- Tests pass.
- New tests cover additional CLI options.

maxwellpeterson-wf:
QA +1
- [x] Dev +1's
  - https://github.com/Workiva/dependency_validator/pull/15#issuecomment-368906834
- [x] Tests created/updated 
- [x] All tests pass

PR #17, opened 2018-02-28
UIP-3088 Fix detection of packages whose names contain numbers

greglittlefield-wf:
## Ultimate Problem
The regex that detected import/export statements didn't account for packages with numbers in their names.

This can cause false positives and false negatives when using dependnency_validator (I hit one of each when using it).

## Solution
- Account for numbers in package names
- Add tests for package name regex, including regression tests for package names with numbers
- Boyscouting: update regex for two additional edge cases:
    - Leading spaces
    - Triple-quoted exports (apparently you can do that)
    - imports/exports not separated by newlines
        - This is definitely an edge case, so let me know if my changes to the regex to support this too severe
    - Note: it looks like adjacent string concatenation is also allowed; supporting that is probably best left for another time (if at all), and best for a tool like `analyzer` as opposed to regex.

## Testing
Verify tests pass

## Areas of Regression
Detection of packages.

### DIFF
--- a/test/import_export_regex_test.dart
+++ b/test/import_export_regex_test.dart
@@ -60,6 +60,18 @@ main() {
         test('with extra whitespace in the line', () {
           sharedTest('   $importOrExport   "package:foo/bar.dart"   ;   ', importOrExport, 'foo');
         });
+
+        test('with multiple ${importOrExport}s in the same line', () {
+          final input = '$importOrExport "package:foo/bar.dart"; $importOrExport "package:bar/foo.dart";';
+
+          expect(input, matches(importExportPackageRegex));
+
+          final allMatches = importExportPackageRegex.allMatches(input).toList();
+          expect(allMatches, hasLength(2));
+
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Module Findings
No security relevant content detected. Please review for security relevance and request security review as needed.

test/import_export_regex_test.dart
```diff
@@ -0,0 +1,78 @@
+// Copyright 2017 Workiva Inc.
```
toddbeckman-wf: 2018

greglittlefield-wf: This is present in all the files; I'll make a separate PR following this to update them if that's cool.

toddbeckman-wf:
+10

- [x] Tests exercise regex
- [x] CI passes

maxwellpeterson-wf:
QA +1
- [x] Dev +1's
  - https://github.com/Workiva/dependency_validator/pull/17#issuecomment-369408078
- [x] Tests created/updated 
- [x] All tests pass

PR #26, opened 2018-05-11
AF-159 validate scss deps

toddbeckman-wf:
# Have no fear, most of this diff is `pubspec.lock` files

## Problem

SCSS files have import statements.

## Solution

Check the packages in those import statements for dependencies to validate.

## Testing

- [ ] CI does the thing
- [ ] Create an SCSS file (or point an existing package to this one and edit an existing SCSS file) inside the `lib` directory and add an import statement `@import 'somepackage:foo';` The imported file doesn't need to exist; all that matters is the package name.
  - [ ] Check that missing packages are called out as missing from `dependencies`.
  - [ ] Check that underpromoted packages are called out (the dep is in `dev_dependencies` but should be in `dependencies`)
- [ ] Do the same for an SCSS file that is in some other directory (such as `test/`).
  - [ ] Check that missing packages are called out as missing from `dev_dependencies`.
  - [ ] Check that overpromoted packages are called out (the dep is in `dependencies` but should be in `dev_dependencies`)

fyi @evanweible-wf @greglittlefield-wf


aviary-wf:
## Security Insights

## Action Items
* Verify [`aviary.yaml`](https://dev.workiva.net/docs/information-security/aviary#configuring-aviary) coverage of security relevant code
---
_Questions or Comments? Reach out on HipChat: InfoSec Forum._

test_fixtures/missing/pubspec.lock
```diff
@@ -0,0 +1,59 @@
+# Generated by pub
```
toddbeckman-wf: Note: These `pubspec.lock` files need to be present in order to run test fixtures, but we don't need to keep the actual versions up to date. Only the names of the dependencies matter.

lib/dependency_validator.dart
```diff
@@ -60,22 +60,37 @@ void run({
...
-    final matches = importExportPackageRegex.allMatches(file.readAsStringSync());
+    final matches = importExportDartPackageRegex.allMatches(file.readAsStringSync());
     for (final match in matches) {
       packagesUsedInPublicFiles.add(match.group(2));
     }
   }
+  for (final file in publicScssFiles) {
+    final matches = importExportScssPackageRegex.allMatches(file.readAsStringSync());
+    for (final match in matches) {
+      packagesUsedInPublicFiles.add(match.group(1)); // TODO: why is this 1?
```
evanweible-wf:
```
\b(import|export)\s+['"]{1,3}package:([a-zA-Z0-9_]+)\/[^;]+
```

The Dart regex supports both imports and exports via `(import|export)`. This is technically a group even though we don't care about which one matched. This is why line 84 above uses `match.group(2)`.

---

```
\@import\s+['"]{1,3}package:\s*([a-zA-Z0-9_]+)\/[^;]+
```

The SCSS regex only needs to look for `@import` statements, so it doesn't need that first group, which is why this line checks `match.group(1)`.

toddbeckman-wf: Ahh that makes sense.

lib/dependency_validator.dart
```diff
@@ -60,22 +60,37 @@ void run({
...
+  // Recursively list all Dart and Scss files in lib/
   final publicDartFiles = <File>[]
     ..addAll(listDartFilesIn('lib/', excludedDirs))
     ..addAll(listDartFilesIn('bin/', excludedDirs));
-  logger.fine('public facing dart files:\n'
-      '${bulletItems(publicDartFiles.map((f) => f.path))}\n');
+
+  final publicScssFiles = <File>[]
+    ..addAll(listScssFilesIn('lib/', excludedDirs))
+    ..addAll(listDartFilesIn('bin/', excludedDirs));
```
greglittlefield-wf: This should be `listScssFilesIn`, not `listDartFilesIn`

lib/src/utils.dart
```diff
@@ -46,13 +49,27 @@ String bulletItems(Iterable<String> items) => items.map((l) => '  * $l').join('\
...
+///
+/// This also excludes Scss files that are in a `packages/` subdirectory.
+Iterable<File> listScssFilesIn(String dirPath, List<String> excludedDirs) =>
+    listTypedFilesIn(dirPath, excludedDirs, 'scss');
+
+/// Returns an iterable of all files ending in .[type] in the given
+/// [dirPath] excluding any sub-directories specified in [excludedDirs].
+///
+/// This also excludes files that are in a `packages/` subdirectory.
+Iterable<File> listTypedFilesIn(String dirPath, List<String> excludedDirs, String type) {
```
greglittlefield-wf: #nit I think `extension` would be a less ambiguous name than `type`.

test/import_export_regex_test.dart
```diff
@@ -21,8 +21,8 @@ import 'package:dependency_validator/src/utils.dart';
 void main() {
   group('importExportPackageRegex matches correctly for', () {
```
greglittlefield-wf: I think we should have similar tests for the SCSS regex.

test_fixtures/missing/pubspec.lock
greglittlefield-wf: Should we exclude them from the `.gitignore`, then, and add this information as a comment?

test/import_export_regex_test.dart
toddbeckman-wf: Nice catch. These tests aren't actually running in CI. Time to add them.

lib/src/utils.dart
```diff
@@ -46,13 +49,27 @@ String bulletItems(Iterable<String> items) => items.map((l) => '  * $l').join('\
...
+///
+/// This also excludes Scss files that are in a `packages/` subdirectory.
+Iterable<File> listScssFilesIn(String dirPath, List<String> excludedDirs) =>
+    listFilesWithExtensionIn(dirPath, excludedDirs, 'scss');
+
+/// Returns an iterable of all files ending in .[type] in the given
+/// [dirPath] excluding any sub-directories specified in [excludedDirs].
+///
+/// This also excludes files that are in a `packages/` subdirectory.
+Iterable<File> listFilesWithExtensionIn(String dirPath, List<String> excludedDirs, String type) {
```
evanweible-wf: s/type/extension

evanweible-wf:
+10
- [x] CI does the thing
- [x] Create an SCSS file (or point an existing package to this one and edit an existing SCSS file) inside the `lib` directory and add an import statement `@import 'package:foo/test';` The imported file doesn't need to exist; all that matters is the package name.
  - [x] Check that missing packages are called out as missing from `dependencies`.
  - [x] Check that underpromoted packages are called out (the dep is in `dev_dependencies` but should be in `dependencies`)
- [x] Do the same for an SCSS file that is in some other directory (such as `test/`).
  - [x] Check that missing packages are called out as missing from `dev_dependencies`.
  - [x] Check that overpromoted packages are called out (the dep is in `dependencies` but should be in `dev_dependencies`)

QA +1
- [x] Dev +1s
- [x] +10 tested
- [x] Rosie is happy

@Workiva/release-management-p

PR #27, opened 2018-05-18
AF-178 check for dep pins

toddbeckman-wf:
## Problem

Dependency pins are bad

## Solution

Check for pins.

## Testing

- [ ] Change a dependency to a pin and run `pub run dependency_validator -i coverage,dartdoc,dart_style,_foo,foo,_bar,bar,foo1,foo_foo` (which is run by task-runner) and verify that `echo $?` returns 1.
- [ ] Repeat with the `--no-fatal-pins` flag. You should still get a warning, but the return code is 0.

Example pins:

* `path: 1.4.2`
* `args: '>=0.13.7 <=2.0.0'`
* `args: '>=0.13.7 <2.0.1'`


lib/src/utils.dart
```diff
@@ -82,3 +83,73 @@ void logDependencyInfractions(String infraction, Iterable<String> dependencies)
...
+    return false;
+  }
+
+  // Direct pin
+  if (constraint is Version) {
+    return true;
+  }
+
+  if (constraint is VersionRange) {
+    // Defined maximum bound is a pin
```
greglittlefield-wf: Just thought... it may not be immediately obvious to consumers why this is considered a pin. 

It may be helpful to make the failure message be more verbose, and outline all the cases that are considered pins and why.

lib/src/utils.dart
```diff
@@ -82,3 +83,73 @@ void logDependencyInfractions(String infraction, Iterable<String> dependencies)
...
+    }
+
+    // Check legal upper bound doesn't block minor bumps
+    if (max.major == 0) {
+      if (max.minor == 0) {
+        return true; // 0.0.x
+      }
+      return max.patch != 0; // 0.x.0
+    }
+    return max.minor + max.patch > 0; // x.0.0
```
greglittlefield-wf: Is this just a tricky way to see whether either is non-zero? If so, I think the code would read more clearly if it was straightforward:
```dart
return max.minor != 0 || max.patch != 0;
```
 If not, what's going on?

lib/src/utils.dart
```diff
@@ -82,3 +83,73 @@ void logDependencyInfractions(String infraction, Iterable<String> dependencies)
...
+    // Builds and prereleases also block patch and minor bumps
+    if (max.build.isNotEmpty ||
+        // First preRelease here is okay because the upper bound is not inclusive
+        (max.isPreRelease && !max.isFirstPreRelease)) {
+      return true;
+    }
+
+    // Check legal upper bound doesn't block minor bumps
+    if (max.major == 0) {
+      if (max.minor == 0) {
```
greglittlefield-wf: Was it determined that 0.0.x versions are a special case and not treated the same 0.x.x?

From the pub semver README, it seems that it's not a special case:
>pub defines the "next breaking" version as the version which increments the major version if it's greater than zero, and the minor version otherwise

test/unit/src/utils_test.dart
```diff
@@ -128,4 +128,136 @@ void main() {
...
+      expect(doesVersionPinDependency('"any"'), isFalse);
+    });
+
+    test('is true for empty', () {
+      expect(doesVersionPinDependency('>0.0.0 <0.0.0'), isTrue);
+      expect(doesVersionPinDependency('">0.0.0 <0.0.0"'), isTrue);
+      expect(doesVersionPinDependency("'>0.0.0 <0.0.0'"), isTrue);
+    });
+
+    test('with carat notation is false', () {
```
greglittlefield-wf: typo: carat/caret

test/unit/src/utils_test.dart
```diff
@@ -128,4 +128,136 @@ void main() {
...
+        expect(doesVersionPinDependency('">=0.2.3 <0.5.6"'), isTrue);
+        expect(doesVersionPinDependency("'>=0.2.3 <0.5.6'"), isTrue);
+        expect(doesVersionPinDependency('"<0.5.6"'), isTrue);
+        expect(doesVersionPinDependency("'<0.5.6'"), isTrue);
+      });
+
+      test('when version starts with nonzero', () {
+        // blocks minor updates
+        expect(doesVersionPinDependency('">=1.2.3 <4.5.0"'), isTrue);
+        expect(doesVersionPinDependency("'>=1.2.3 <4.5.0'"), isTrue);
```
greglittlefield-wf: For future-proofing changes to this logic, could we add test cases for blocking minor/patch updates for the same major/minor versions?
```dart
expect(doesVersionPinDependency("'>=1.2.3 <1.2.4'"), isTrue);
expect(doesVersionPinDependency("'>=1.3.0 <1.4.0'"), isTrue);
```

lib/src/utils.dart
```diff
@@ -82,3 +83,73 @@ void logDependencyInfractions(String infraction, Iterable<String> dependencies)
...
+      final Map packageMeta = dependencies[packageName];
+      if (packageMeta.containsKey('version')) {
+        version = packageMeta['version'];
+      } else {
+        // This feature only works for versions, not git refs or paths.
+        break;
+      }
+    }
+
+    if (doesVersionPinDependency(version)) {
```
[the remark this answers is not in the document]
toddbeckman-wf: I'm not sure how pub get would resolve with a null version.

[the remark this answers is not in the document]
greglittlefield-wf: Oh wait, I see that this is inside the for-loop, so the break should handle that...

Wait, would the break prevent subsequent dependencies from being validated?

lib/src/utils.dart
toddbeckman-wf: Yeah it was just code golfing. I can change it to more readable.

lib/src/utils.dart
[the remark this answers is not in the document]
toddbeckman-wf: Nice catch <_<

lib/src/utils.dart
greglittlefield-wf: Thanks! I appreciate the code golfing spirit, though 😄

lib/src/constants.dart
```diff
@@ -0,0 +1,64 @@
...
+/// String key in pubspec.yaml for the dev_dependencies map.
+const String devDependenciesKey = 'dev_dependencies';
+
+/// String key in pubspec.yaml for the package name.
+const String nameKey = 'name';
+
+/// String key in pubspec.yaml for the transformers map.
+const String transformersKey = 'transformers';
+
+/// Provides a set of reasons why versions are ppins and whether
```
greglittlefield-wf: Looks like this comment never finished getting typed 😄

lib/src/constants.dart
```diff
@@ -0,0 +1,64 @@
...
+  /// Whether this is a pin.
+  final bool isPin;
+
+  DependencyPinEvaluation._(this.message, {this.isPin: true});
+
+  @override
+  String toString() => message;
+
+  /// <1.2.0
+  static final DependencyPinEvaluation blocksMinorBumps = new DependencyPinEvaluation._('This pin blocks minor bumps.');
```
greglittlefield-wf: #nit These could be `const`s instead.

lib/src/constants.dart
```diff
@@ -0,0 +1,64 @@
...
+
+  /// >1.2.3 <1.2.3
+  static final DependencyPinEvaluation emptyPin =
+      new DependencyPinEvaluation._('Empty dependency versions cannot be resolved.');
+
+  /// <=1.2.3
+  static final DependencyPinEvaluation inclusiveMax =
+      new DependencyPinEvaluation._('Inclusive max bounds restrict minor bumps and patches.');
+
+  /// :)
```
greglittlefield-wf: 😄

lib/src/utils.dart
```diff
@@ -89,67 +69,68 @@ List<String> getDependenciesWithPins(Map dependencies) {
...
+        continue;
       }
     }
 
-    if (doesVersionPinDependency(version)) {
-      infractions.add(packageName);
+    final DependencyPinEvaluation evaluation = inspectVersionForPins(version);
+
+    if (evaluation.isPin) {
+      infractions.add('$packageName -- ${evaluation.message}');
```
greglittlefield-wf: Thoughts on also including the bad version range here, for context?

toddbeckman-wf: I figured they already have the pubspec.yaml

lib/src/utils.dart
```diff
@@ -82,3 +63,74 @@ void logDependencyInfractions(String infraction, Iterable<String> dependencies)
...
+/// Lists the packages with infractions
+List<String> getDependenciesWithPins(Map dependencies) {
+  final List<String> infractions = [];
+  for (String packageName in dependencies.keys) {
+    String version;
+    final packageMeta = dependencies[packageName];
+
+    if (dependencies[packageName] is String) {
+      version = packageMeta;
+    } else {
```
greglittlefield-wf: Looks like this didn't get addressed:

> Should we have a check for `is Map` here? The Yaml spec is pretty complex and their may be some edge-cases here.

I think we should do this. I actually just ran into it when testing wdesk_sdk (lolwut):
https://github.com/Workiva/wdesk_sdk/blob/master/pubspec.yaml#L38

toddbeckman-wf: Does empty string resolve as any?

greglittlefield-wf:
Hm, question, how should we handle pins that are in place to help promote faster `pub get` version solving by preventing minor/patch upgrades to versions that are known to be Dart2-only?

Example:
```
Validating dependencies for over_react

These packages are pinned in pubspec.yaml:
  * built_value: >=4.2.0 <5.2.0 -- This pin blocks minor bumps.
```

They could add `--no-fatal-pins` to work around it, but its a bummer that they don't get failures for any other pins. Would it be worth being able to ignore these? (I'm not advocating for that, just posing the question.)

toddbeckman-wf:
We've already had really weird `built_*` version resolution issues, and I don't know whether it's okay to pin the max bound for resolution time if it comes at the cost of blocking dependencies.

However, I think if we know there are specific packages that are problematic for resolution, we could potentially `--ignore` them. But I don't have strong feelings on that and can be persuaded.

lib/src/utils.dart
```diff
@@ -82,3 +63,76 @@ void logDependencyInfractions(String infraction, Iterable<String> dependencies)
...
+        continue;
+      }
+    } else {
+      continue; // no version string set
+    }
+
+    final DependencyPinEvaluation evaluation = inspectVersionForPins(version);
+
+    if (evaluation.isPin) {
+      infractions.add('$packageName -- ${evaluation.message}');
```
greglittlefield-wf: I still think it would be nice to add the version range here since we have it so handy: 
```dart
'$packageName: $version -- ${evaluation.message}'
```

If there's multiple, it reduces the need to cross-reference them, especially if they're printed in CI or something:
```
These packages are pinned in pubspec.yaml:
  * analyzer -- Inclusive max bounds restrict minor bumps and patches.
  * barback -- Inclusive max bounds restrict minor bumps and patches.
  * built_value -- This pin blocks minor bumps.
  * quiver -- Inclusive max bounds restrict minor bumps and patches.
```
vs
```
These packages are pinned in pubspec.yaml:
  * analyzer: >=0.30.0 <=0.31.0 -- Inclusive max bounds restrict minor bumps and patches.
  * barback: >=0.15.2 <=0.15.2+14 -- Inclusive max bounds restrict minor bumps and patches.
  * built_value: >=4.2.0 <5.2.0 -- This pin blocks minor bumps.
  * quiver: >=0.21.4 <=0.28.0 -- Inclusive max bounds restrict minor bumps and patches.
```

Not a must have, but I'd personally like to have it 😸

bin/dependency_validator.dart
```diff
@@ -22,6 +22,7 @@ final ArgParser argParser = new ArgParser()
   ..addFlag('verbose', defaultsTo: false)
   ..addOption('ignore', abbr: 'i', allowMultiple: true, splitCommas: true)
   ..addOption('exclude-dir', abbr: 'x', allowMultiple: true, splitCommas: true)
+  ..addFlag('fatal-pins', defaultsTo: true)
```
greglittlefield-wf: This isn't documented anywhere, and can't be discovered by consumers without reading the source code.

This should be added to the readme. This might also be a good opportunity to fold in work to do the `--help` flag: https://jira.atl.workiva.net/browse/AF-137, if you're feeling ambitious!

greglittlefield-wf:
I don't have strong opinions either, so I'll defer to other reviewers on that.

bin/dependency_validator.dart
toddbeckman-wf: I'd rather do that in a follow-up PR just for separation of concerns.

greglittlefield-wf: I'm fine with that, so long as we do it soon, before consumers need to figure out how to use the flag. If it's going to take a while to get to, we should just stick a line in the README.

toddbeckman-wf: On second thought I added it here: https://github.com/Workiva/dependency_validator/pull/28

I can change the base of that PR to this one and merge it directly here if you'd like.

maxwellpeterson-wf:
QA +1
- [x] Dev +1's
  - https://github.com/Workiva/dependency_validator/pull/27#pullrequestreview-123211214
- [x] Tests created/updated 
- [x] All tests pass

PR #50, opened 2019-09-12
CPLAT-7378: Add build_vm_compilers to common binary packages.

evanweible-wf:
# [CPLAT-7378](https://jira.atl.workiva.net/browse/CPLAT-7378)
Issue Status

## Motivation
  
- `build_vm_compilers` may also become a commonly used binary package that we can ignore by default

## Changes
  
- add it to the list!


evanweible-wf:
QA +1
- [x] Verified locally in a project with build_vm_compilers and coverage that both are ignored by default

@Workiva/release-management-p

PR #56, opened 2020-01-09
CPLAT-10684 Spike out support for static configuration in pubspec.yaml

evanweible-wf:
## Motivation
Support static configuration of this tool within a project so that the executable can be run without any arguments. In other words, you shouldn't need to know anything about a project in order to run this validation.


aviary3-wk:
## Security Insights

## Action Items
---

lib/dependency_validator.dart
```diff
@@ -14,6 +14,8 @@
 
 import 'dart:io';
 
+import 'package:dependency_validator/src/pubspec_config.dart';
```
todbachman-wf: Any reason this isn't a path import?

lib/dependency_validator.dart
```diff
@@ -43,38 +55,29 @@ void run({
...
-  final publicDartFiles = <File>[]
-    ..addAll(listDartFilesIn('lib/', excludedDirs))
-    ..addAll(listDartFilesIn('bin/', excludedDirs));
-
-  final publicScssFiles = <File>[]
-    ..addAll(listScssFilesIn('lib/', excludedDirs))
-    ..addAll(listScssFilesIn('bin/', excludedDirs));
+  final publicDirs = ['bin/', 'lib/'];
+  final publicDartFiles = [
+    for (final dir in publicDirs) ...listDartFilesIn(dir, excludes),
```
todbachman-wf: Sweet spread syntax!

lib/dependency_validator.dart
```diff
@@ -23,15 +25,25 @@ export 'src/constants.dart' show commonBinaryPackages;
...
 void run({
-  List<String> excludedDirs = const [],
-  bool fatalDevMissing = true,
-  bool fatalMissing = true,
-  bool fatalOverPromoted = true,
-  bool fatalPins = true,
-  bool fatalUnderPromoted = true,
-  bool fatalUnused = true,
-  List<String> ignoredPackages = const [],
+  List<Glob> excludes,
```
evanweible-wf: this is technically a breaking change (I replaced `excludedDirs` with `excludes`), but if we want to release this as a minor change to start, we could keep `excludedDirs` and deprecate it

lib/dependency_validator.dart
```diff
@@ -23,15 +25,25 @@ export 'src/constants.dart' show commonBinaryPackages;
...
+  bool fatalPins,
+  bool fatalUnderPromoted,
+  bool fatalUnused,
+  List<String> ignoredPackages,
 }) {
+  final config = PubspecDepValidatorConfig.fromYaml(
+          File('pubspec.yaml').readAsStringSync())
+      .dependencyValidator;
+  excludes ??= [];
+  excludes.addAll(config?.exclude?.map((s) => Glob(s)) ?? []);
```
evanweible-wf, line 41: We should wrap this in a try-catch in case one of the string values in the yaml config is not a valid Glob

lib/dependency_validator.dart
```diff
@@ -108,61 +106,59 @@ void run({
...
-    ..addAll(listScssFilesIn('web/', excludedDirs));
-  final nonLibLessFiles = <File>[]
-    ..addAll(listLessFilesIn('example/', excludedDirs))
-    ..addAll(listLessFilesIn('test/', excludedDirs))
-    ..addAll(listLessFilesIn('tool/', excludedDirs))
-    ..addAll(listLessFilesIn('web/', excludedDirs));
+  final nonPublicDartFiles = listDartFilesIn('./', [
+    ...excludes,
+    for (final dir in publicDirs) Glob('$dir**'),
+  ]);
```
evanweible-wf, line 115: [nit] this list of exclude globs here is the same for the next two usages, as well. Could move it to a variable to DRY this up.

pubspec.yaml
```diff
@@ -10,15 +10,33 @@ environment:
...
+  exclude:
+    - 'test_fixtures/**'
+  ignore:
+    - json_serializable
+    - _foo
+    - foo
+    - _bar
+    - bar
+    - foo1
+    - foo_foo
```
evanweible-wf, line 42: note to self: these packages only have to be ignored because our regex approach is picking up on imports in strings and not real imports. We should consider using the `analyzer` package and an `AstVisitor` to detect imports and exports in Dart code; it should still be relatively fast since it only needs to parse the AST and not analyze the files.

pubspec.yaml
```diff
@@ -10,15 +10,33 @@ environment:
...
+  workiva_analysis_options: ^1.0.0
 
 executables:
   dependency_validator:
+
+dependency_validator:
+  exclude:
+    - 'test_fixtures/**'
+  ignore:
+    - json_serializable
```
evanweible-wf, line 36: note to self: we should look up dependencies in the pub-cache to see if they provide a `build.yaml` that defines a builder - if that's the case, it's likely that the dependency is needed for its builder and shouldn't need to be ignored.

lib/dependency_validator.dart
```diff
@@ -22,16 +24,24 @@ import 'src/utils.dart';
...
-  bool fatalDevMissing = true,
-  bool fatalMissing = true,
-  bool fatalOverPromoted = true,
-  bool fatalPins = true,
-  bool fatalUnderPromoted = true,
-  bool fatalUnused = true,
-  List<String> ignoredPackages = const [],
-}) {
+void run() {
+  final config = PubspecDepValidatorConfig.fromYaml(File('pubspec.yaml').readAsStringSync()).dependencyValidator;
```
evanweible-wf, line 28: We should check to see if `File('pubspec.yaml')` exists and display a more helpful error message if it doesn't

test/executable_test.dart
```diff
@@ -76,41 +33,110 @@ ProcessResult checkProject(
...
+      // Create fake project that any test may use
+      final fakeProjectPubspec = ''
+          'name: fake_project\n'
+          'version: 0.0.0\n'
+          'private: true\n'
+          'environment:\n'
+          '  sdk: \'>=2.4.0 <3.0.0\'\n'
+          'dev_dependencies:\n'
+          '  dependency_validator:\n'
+          '    path: ${Directory.current.path}\n';
```
evanweible-wf, line 46: To avoid having to escape quotes and manually insert the `\n` line breaks, we could copy the `unindent()` util:
https://github.com/Workiva/over_react/blob/315f62be4a8462d841dea31b6563cfb976fa71f3/lib/src/util/string_util.dart#L17-L44

and then use multiline strings like so:

```dart
      final fakeProjectPubspec = unindent('''
          name: fake_project
          version: 0.0.0
          private: true
          environment:
            sdk: '>=2.4.0 <3.0.0'
          dev_dependencies:
            dependency_validator:
              path: ${Directory.current.path}
          ''');
```

test/executable_test.dart
```diff
@@ -13,60 +13,17 @@
 // limitations under the License.
 
 @TestOn('vm')
-
 import 'dart:io';
 
-import 'package:dependency_validator/dependency_validator.dart';
 import 'package:test/test.dart';
+import 'package:test_descriptor/test_descriptor.dart';
```
evanweible-wf, line 19: It's a bit weird, but the convention for importing and using this package is to import it with a prefix of `d`. So far we've followed that convention in the few packages at workiva that use it so I think it'd be good to be consistent.

test/executable_test.dart
aaronstgeorge-wf: Ooooh I was looking for something like that!

PR #58, opened 2020-04-01
Validate less imports

Anrock:
## Motivation
Take imports from .less file into account to produce more accurate dependency analysis

## Changes
Read and analyze .less files for imports. Basically same as scss but different file extension and regex.

#### Release Notes
Support .less files analysis

## Review
Please review: @Workviva/app-frameworks

### QA Checklist
- [ ] Tests were updated and provide good coverage of the changeset and other affected code
- [ ] Manual testing was performed if needed
- [ ] Anything falling under manual testing criteria [outlined in CONTRIBUTING.md][contributing-manual-testing]

## Merge Checklist
While we perform many automated checks before auto-merging, some manual checks are needed:
- [ ] A Client Platform member has reviewed these changes
- [ ] There are no unaddressed comments _- this check can be automated if reviewers use the "Request Changes" feature_
- [ ] _For release PRs -_ Version metadata in Rosie comment is correct

PR #59, opened 2020-04-03
CPLAT-10392: v1.5.0

evanweible-wf:
Changes from #58 with an additional commit to update the changelog.

### DIFF
--- a/lib/src/constants.dart
+++ b/lib/src/constants.dart
@@ -6,8 +6,7 @@ final RegExp importExportDartPackageRegex =
 final RegExp importScssPackageRegex = RegExp(r'''\@import\s+['"]{1,3}package:\s*([a-zA-Z0-9_]+)\/[^;]+''');
 
 /// Regex used to detect all Less import directives.
-final RegExp importLessPackageRegex =
-    RegExp(r'@import[\t ]+(?:\(.*\)\s+)?"(?:packages\/|package:\/\/)([a-zA-Z1-9_-]+)\/');
+final RegExp importLessPackageRegex = RegExp(r'@import\s+(?:\(.*\)\s+)?"(?:packages\/|package:\/\/)([a-zA-Z1-9_-]+)\/');
 
 /// String key in pubspec.yaml for the dependencies map.
 const String dependenciesKey = 'dependencies';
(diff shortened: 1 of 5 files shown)


aviary-wf:
## Security Insights

## Action Items
---

lib/src/constants.dart
```diff
@@ -5,6 +5,10 @@ final RegExp importExportDartPackageRegex =
 /// Regex used to detect all Sass import directives.
 final RegExp importScssPackageRegex = RegExp(r'''\@import\s+['"]{1,3}package:\s*([a-zA-Z0-9_]+)\/[^;]+''');
 
+/// Regex used to detect all Less import directives.
+final RegExp importLessPackageRegex =
+    RegExp(r'@import[\t ]+(?:\(.*\)\s+)?"(?:packages\/|package:\/\/)([a-zA-Z1-9_-]+)\/');
```
seanburke-wf, line 10: Is there a reason to use `[\t ]+` after `@import` instead of `\s+`?

test/utils_test.dart
```diff
@@ -104,6 +104,52 @@ include: package:pedantic/analysis_options.1.8.0.yaml
...
+        sharedTest('@import "package://_foo/bar";', '_foo');
+      });
+    });
+
+    test('with extra whitespace in the line', () {
+      sharedTest('   @import   "packages/foo/bar"   ;   ', 'foo');
+      sharedTest('   @import   "package://foo/bar"   ;   ', 'foo');
+    });
+
+    test('with multiple import\'s in the same line', () {
```
seanburke-wf, line 140: #nit: "import's" -> "imports"

lib/src/constants.dart
evanweible-wf: I'm not familiar with LESS so I'll defer to @Anrock

Anrock: I'm not really familiar with less either, but from the spec it seemed like no newlines were allowed after `@import` and `s` includes newlines. So I've used `[\t ]` to allow only non-linebreaking whitespace.

seanburke-wf: My biggest concern here was other whitespace (non-breaking space or zero-width, for example), but playing with http://lesscss.org/less-preview/ it does look like newline is allowed.

evanweible-wf:
@seanburke-wf addressed your comments.

evanweible-wf:
QA +1
@Workiva/release-management-p

PR #61, opened 2020-06-08
CPLAT-11300 detect packages that provide one or more executables

aaronstgeorge-wf:
## Motivation
  

There are several popular packages that currently have to be ignored when running dependency_validator because they are only used for the executable(s) they provide. The usage of these executables cannot be reliably detected by scanning project files because they can be run in many different ways, or they may only be run locally and not referenced in code.

## Changes
  

Look up the source for each dependency in ~/.pub-cache to see a bin/ directory exists and contains exectuable entrypoints. If it does, we can take a conservative approach and refuse to consider that dependency "unused" as its executable(s) could be used.

## Review
_[See CONTRIBUTING.md][contributing-review-types] for more details on review types (+1 / QA +1 / +10) and code review process._

  

Please review: 

### QA Checklist
- [ ] Tests were updated and provide good coverage of the changeset and other affected code
- [ ] Manual testing was performed if needed
    - [ ] Steps from PR author: 
        
    - [ ] Anything falling under manual testing criteria [outlined in CONTRIBUTING.md][contributing-manual-testing]

## Merge Checklist
While we perform many automated checks before auto-merging, some manual checks are needed:
- [ ] A Client Platform member has reviewed these changes
- [ ] There are no unaddressed comments _- this check can be automated if reviewers use the "Request Changes" feature_
- [ ] _For release PRs -_ Version metadata in Rosie comment is correct

[contributing-review-types]: https://github.com/Workiva/dependency_validator/blob/master/CONTRIBUTING.md#review-types
[contributing-manual-testing]: https://github.com/Workiva/dependency_validator/blob/master/CONTRIBUTING.md#manual-testing-criteria

### DIFF
--- a/lib/dependency_validator.dart
+++ b/lib/dependency_validator.dart
@@ -16,6 +16,7 @@ import 'dart:io';
 import 'dart:convert';
 
 import 'package:glob/glob.dart';
+import 'package:package_config/package_config.dart';
 import 'package:yaml/yaml.dart';
 
 import 'src/constants.dart';
@@ -25,7 +26,7 @@ import 'src/utils.dart';
 export 'src/constants.dart' show commonBinaryPackages;
 
 /// Check for missing, under-promoted, over-promoted, and unused dependencies.
-void run() {
+Future<Null> run() async {
   if (!File('pubspec.yaml').existsSync()) {
     logger.shout('pubspec.yaml not found');
     exit(1);
@@ -34,27 +35,6 @@ void run() {
     logger.shout('.dart_tool/package_config.json not found');
     exit(1);
   }
-
-  // TODO: maybe we can get the actual deserialization code for this file from the builder package
-  final packageConfig = jsonDecode(File('.dart_tool/package_config.json').readAsStringSync());
-
-  // TODO: document why we are doing this
-  final rootDir = Directory.current;
-  final dir = Directory('${rootDir.path}/.dart_tool');
-  Directory.current = dir;
-
-  final packagesWithExecutables = Set<String>();
-  for (final thing in packageConfig['packages']) {
-    final uri = thing['rootUri'] as String;
-    final path = uri.replaceFirst('file://', '');
-    if (Directory('$path/bin').existsSync()) {
-      packagesWithExecutables.add(thing['name']);
-    }
-  }
-
-  // TODO: explain.
-  // TODO: Maybe there is a directory stack already around in dart?
-  Directory.current = rootDir;
 
   final config = PubspecDepValidatorConfig.fromYaml(File('pubspec.yaml').readAsStringSync()).dependencyValidator;
   final configExcludes = config?.exclude
@@ -83,6 +63,26 @@ void run() {
   final packageName = pubspecYaml[nameKey];
 
   logger.info('Validating dependencies for $packageName\n');
+
+  // `loadPackageConfig` (used below) relies on having a file built from an absolute URI.
+  final packageConfig = File.fromUri(Uri.file('${Directory.current.path}/.dart_tool/package_config.json'));
+
+  // URIs in packageConfig may be relative to its directory. Change the working directory to account for this.
+  final rootDir = Directory.current;
+  final dir = Directory('${rootDir.path}/.dart_tool');
+  Directory.current = dir;
+
+  // Find packages that provide executables.
+  final packagesWithExecutables = Set<String>();
+  for (final package in (await loadPackageConfig(packageConfig)).packages) {
+    final path = package.root.path;
+    if (Directory('$path${path.endsWith('/') ? '' : '/'}bin').existsSync()) {
+      packagesWithExecutables.add(package.name);
+    }
+  }
+
+  // Set working directory back.
+  Directory.current = rootDir;
 
   logDependencyInfo(
       'the following packages contain executables, they are assumed to be used:', packagesWithExecutables);
--- a/lib/dependency_validator.dart
+++ b/lib/dependency_validator.dart
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import 'dart:io';
-import 'dart:convert';
 
 import 'package:glob/glob.dart';
 import 'package:package_config/package_config.dart';
@@ -267,7 +266,7 @@ Future<Null> run() async {
       // Start with all explicitly declared dependencies
       deps
           .union(devDeps)
-          // Remove all deps that provide and executable
+          // Remove all deps that provide an executable
           .difference(packagesWithExecutables)
           // Remove all deps that were used in Dart code somewhere in this package
           .difference(packagesUsedInPublicFiles)
--- a/lib/dependency_validator.dart
+++ b/lib/dependency_validator.dart
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import 'dart:io';
+import 'dart:convert';
 
 import 'package:glob/glob.dart';
 import 'package:yaml/yaml.dart';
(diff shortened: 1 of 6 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/dependency_validator.dart
```diff
@@ -24,11 +25,16 @@ import 'src/utils.dart';
...
-void run() {
+Future<Null> run() async {
   if (!File('pubspec.yaml').existsSync()) {
     logger.shout('pubspec.yaml not found');
     exit(1);
   }
+  if (!File('.dart_tool/package_config.json').existsSync()) {
+    logger.shout('.dart_tool/package_config.json not found');
+    exit(1);
+  }
```
evanweible-wf, line 36: This is effectively a proxy for whether or not `pub get` has been run. In the ticket I was suggesting that we log a warning when this occurs telling the user that some packages may incorrectly marked as unused until a `pub get` is run so that we can do this additional executable checking logic.. but since we're working on a major release, I guess we really just need to decide whether we should make that a requirement. Thoughts?

If we do decide that the existence of the package config file (in other words, having run a `pub get`) is a requirement for running dependency validator, then we should update this error message to be more helpful by telling the user to run `pub get` first.

lib/dependency_validator.dart
```diff
@@ -57,6 +63,29 @@ void run() {
...
+  final packageConfig = File.fromUri(Uri.file('${Directory.current.path}/.dart_tool/package_config.json'));
+
+  // URIs in packageConfig may be relative to its directory. Change the working directory to account for this.
+  final rootDir = Directory.current;
+  final dir = Directory('${rootDir.path}/.dart_tool');
+  Directory.current = dir;
+
+  // Find packages that provide executables.
+  final packagesWithExecutables = Set<String>();
+  for (final package in (await loadPackageConfig(packageConfig)).packages) {
```
evanweible-wf, line 76: There is a separate `package_config_discovery.dart` entry point that provides APIs for discovering and parsing the package config file. We should be able to use [`findPackageConfig(Directory.current)`](https://pub.dev/documentation/package_config/latest/package_config.package_config_discovery/findPackageConfig.html) to replace most of this logic.

lib/dependency_validator.dart
```diff
@@ -57,6 +63,29 @@ void run() {
...
+    final path = package.root.path;
+    if (Directory('$path${path.endsWith('/') ? '' : '/'}bin').existsSync()) {
```
evanweible-wf, line 78: The [`path`](https://pub.dev/packages/path) package should be used here to join filepaths:
```dart
import 'package:path/path.dart' as p;
```

```suggestion
    if (Directory(p.join(package.root.path, 'bin')).existsSync()) {
```

lib/dependency_validator.dart
```diff
@@ -57,6 +63,29 @@ void run() {
...
+      packagesWithExecutables.add(package.name);
+    }
+  }
+
+  // Set working directory back.
+  Directory.current = rootDir;
```
evanweible-wf, line 84: should be able to remove this once `findPackageConfig()` above is used

lib/dependency_validator.dart
```diff
@@ -57,6 +63,29 @@ void run() {
...
+
+  logDependencyInfo(
+      'the following packages contain executables, they are assumed to be used:', packagesWithExecutables);
```
evanweible-wf, line 87: I think this is going to be too noisy since above we're tracking all transitive dependencies and not just those that are directly depended on by this package. In addition to only checking direct dependencies for executables (instead of all transitive ones, as well), I think we should filter this list to be the packages that are being considered "used" _only_ because they contain executables and not for any other reason. That way this info log can be an indicator to maintainers that the listed packages might be able to be removed if their executables aren't being used.

lib/dependency_validator.dart
evanweible-wf, line 76: Instead of checking every package (which includes transitive dependencies), we should move this block down below the code that parses the `pubspec.yaml` so that we can ignore the packages that aren't direct dependencies since they would never show up as "unused" anyway

lib/src/constants.dart
```diff
@@ -22,14 +22,10 @@ const String nameKey = 'name';
...
 /// Packages that are typically only used for their binaries.
 const List<String> commonBinaryPackages = [
-  'build_runner',
   'build_test',
   'build_vm_compilers',
   'build_web_compilers',
   'built_value_generator',
-  'coverage',
-  'dart_dev',
-  'dart_style',
```
evanweible-wf, line 32: :nice:

lib/src/utils.dart
```diff
@@ -92,6 +92,13 @@ void logDependencyInfractions(String infraction, Iterable<String> dependencies)
   logger.warning([infraction, bulletItems(sortedDependencies), ''].join('\n'));
 }
 
+/// Logs a info with the given [info] and lists all of the given
+/// [dependencies] under that.
+void logDependencyInfo(String info, Iterable<String> dependencies) {
+  final sortedDependencies = dependencies.toList()..sort();
+  logger.info([info, bulletItems(sortedDependencies), ''].join('\n'));
+}
```
evanweible-wf, line 100: We could probably just generalize the `logDependencyInfractions()` method above and have it take an optional log `Level`.

aaronstgeorge-wf:
>We'll also want to add a test that verifies that a dependency that is otherwise unused but contains a script in bin/ is not marked as unused. Should be able to do that by manually overwriting the contents of .dart_tool/package.config.json in the test sandbox dir such that one of the root package URIs points to a subdirectory in that test sandbox that is setup to be a package with a bin/

I think the existing test for common binaries will already accomplish this now that packages that provide executables such as `build_runner` and `coverage` are removed.

https://github.com/Workiva/dependency_validator/blob/81267c5a8b9e913c7880791e22192c55759da928/test/executable_test.dart#L339-L370

lib/dependency_validator.dart
aaronstgeorge-wf: >In the ticket I was suggesting that we log a warning when this occurs telling the user that some packages may incorrectly marked as unused

Sorry, that's totally in the ticket description, I didn't read the ticket closely enough :coneofshame:.

>I guess we really just need to decide whether we should make that a requirement. Thoughts?

I'd lean towards requiring a `pub get`. Personally I'd prefer a tool that is persnickety but always gives the same result, to one that gives different answers depending on circumstances.

>If we do decide that the existence of the package config file (in other words, having run a pub get) is a requirement for running dependency validator, then we should update this error message to be more helpful by telling the user to run pub get first.

Roger :salute:.

lib/dependency_validator.dart
aaronstgeorge-wf: Since we are only reporting the intersection between this set and the set of packages explicitly depended on I'm not sure why it matters if there are extra packages in this set?

lib/dependency_validator.dart
aaronstgeorge-wf: I'll hold off on any changes related to this until I hear back from you.

lib/dependency_validator.dart
evanweible-wf: Yeah, if we only report the intersection the end result will be the same, but it does mean we're doing work (checking the source of transitive deps) unnecessarily. Might not actually take that long to do since we're just parsing a couple file paths and checking to see if a directory exists, but `wdesk` for example has 260 total dependencies so it might add up.

lib/dependency_validator.dart
evanweible-wf: Yeah I agree, better to be consistent and a `pub get` is a reasonable bare minimum for running a check on a dart package IMO. I dropped a message in our team chat to see if anyone else has thoughts.

lib/dependency_validator.dart
evanweible-wf: Eh.. looking at the code again, it's so simple that it's probably not worth changing this. Running in wdesk takes ~3.5s and in this repo it takes ~2s so there isn't much juice to squeeze there anyway.

lib/dependency_validator.dart
```diff
@@ -33,7 +33,7 @@ Future<Null> run() async {
     exit(1);
   }
   if (!File('.dart_tool/package_config.json').existsSync()) {
-    logger.shout('.dart_tool/package_config.json not found');
+    logger.shout('No .dart_tool/package_config.json file found, please run "pub get" first.');
```
aaronstgeorge-wf, line 36: This is the same error message webdev gives.

PR #62, opened 2020-06-10
CPLAT-11298 detect builders

aaronstgeorge-wf:
## Motivation
  

There are several popular packages that currently have to be ignored when running dependency_validator because they are only used for the builder they provide, which is something that gets applied automatically when running build_runner and not something we can detect via imports in the source code.

## Changes
  

Look up the source for each dependency in ~/.pub-cache to see if a build.yaml exists, parse it, and see if any builders are defined. If a dependency defines builder(s), we should consider it "used".

## Review
_[See CONTRIBUTING.md][contributing-review-types] for more details on review types (+1 / QA +1 / +10) and code review process._

  

Please review: 

### QA Checklist
- [ ] Tests were updated and provide good coverage of the changeset and other affected code
- [ ] Manual testing was performed if needed
    - [ ] Steps from PR author: 
        
    - [ ] Anything falling under manual testing criteria [outlined in CONTRIBUTING.md][contributing-manual-testing]

## Merge Checklist
While we perform many automated checks before auto-merging, some manual checks are needed:
- [ ] A Client Platform member has reviewed these changes
- [ ] There are no unaddressed comments _- this check can be automated if reviewers use the "Request Changes" feature_
- [ ] _For release PRs -_ Version metadata in Rosie comment is correct

[contributing-review-types]: https://github.com/Workiva/dependency_validator/blob/master/CONTRIBUTING.md#review-types
[contributing-manual-testing]: https://github.com/Workiva/dependency_validator/blob/master/CONTRIBUTING.md#manual-testing-criteria

### DIFF
--- a/lib/dependency_validator.dart
+++ b/lib/dependency_validator.dart
@@ -25,8 +25,6 @@ import 'src/constants.dart';
 import 'src/pubspec_config.dart';
 import 'src/utils.dart';
 
-export 'src/constants.dart' show commonBinaryPackages;
-
 /// Check for missing, under-promoted, over-promoted, and unused dependencies.
 Future<Null> run() async {
   if (!File('pubspec.yaml').existsSync()) {
@@ -59,31 +57,9 @@ Future<Null> run() async {
   final optionsIncludePackage = getAnalysisOptionsIncludePackage();
 
   // Read and parse the pubspec.yaml in the current working directory.
-  final pubspec = Pubspec.parse(File('pubspec.yaml').readAsStringSync());
+  final pubspec = Pubspec.parse(File('pubspec.yaml').readAsStringSync(), sourceUrl: 'pubspec.yaml');
 
-  // Extract the package name.
-  final packageName = pubspec.name;
-
-  logger.info('Validating dependencies for $packageName\n');
-
-  // Find packages that provide executables.
-  final packagesWithExecutables = Set<String>();
-  final packagesWithBuilders = Set<String>();
-  final packageConfig = await findPackageConfig(Directory.current);
-  for (final package in packageConfig.packages) {
-    // Search for builders
-    final buildConfig = await BuildConfig.fromBuildConfigDir(packageName, pubspec.dependencies.keys, package.root.path);
-    if (buildConfig.builderDefinitions.isNotEmpty) {
-      packagesWithBuilders.add(package.name);
-    }
-
-    // Search for executables
-    final binDir = Directory(p.join(package.root.path, 'bin'));
-    hasDartFiles() => binDir.listSync().any((entity) => entity.path.endsWith('.dart'));
-    if (binDir.existsSync() && hasDartFiles()) {
-      packagesWithExecutables.add(package.name);
-    }
-  }
+  logger.info('Validating dependencies for ${pubspec.name}\n');
 
   checkPubspecForPins(pubspec, ignoredPackages: ignoredPackages);
 
@@ -191,7 +167,7 @@ Future<Null> run() async {
           .difference(deps)
           .difference(devDeps)
             // Ignore self-imports - packages have implicit access to themselves.
-            ..remove(packageName)
+            ..remove(pubspec.name)
             // Ignore known missing packages.
             ..removeAll(ignoredPackages);
 
@@ -213,7 +189,7 @@ Future<Null> run() async {
           .difference(devDeps)
           .difference(deps)
             // Ignore self-imports - packages have implicit access to themselves.
-            ..remove(packageName)
+            ..remove(pubspec.name)
             // Ignore known missing packages.
             ..removeAll(ignoredPackages);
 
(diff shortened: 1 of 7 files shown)


lib/dependency_validator.dart
```diff
@@ -51,41 +52,47 @@ Future<Null> run() async {
...
   // Find packages that provide executables.
   final packagesWithExecutables = Set<String>();
+  final packagesWithBuilders = Set<String>();
   final packageConfig = await findPackageConfig(Directory.current);
   for (final package in packageConfig.packages) {
+    // Search for builders
+    final buildConfig = await BuildConfig.fromBuildConfigDir(packageName, pubspec.dependencies.keys, package.root.path);
+    if (buildConfig.builderDefinitions.isNotEmpty) {
+      packagesWithBuilders.add(package.name);
+    }
```
evanweible-wf, line 78: Since we have the fully parsed builder definitions, we should be able to go a little further than just determining whether the dependency defines any builders. It's possible that the builder(s) defined in a package may not be auto-applied–in which case the package might actually be unused–but it's also possible that the root package explicitly enables a builder in its own build.yaml.

Something like this should be able to determine whether the root package has a build.yaml config that references a builder from any given dependency:

```dart
  final rootBuildConfig = await BuildConfig.fromPackageDir('.');
  bool rootPackageReferencesDependencyInBuildYaml(String dependencyName) =>
    [
      ...rootBuildConfig.globalOptions.keys,
      for (final target in rootBuildConfig.buildTargets.values)
        ...target.builders.keys,
    ]
      .map(normalizeBuilderKeyUsage)
      .any((key) => key.startsWith('$dependencyName:');
```

And something like this should be able to determine whether a BuildConfig defines any builders that would be auto-applied to the root package:

```dart
  bool buildConfigDefinesAutoAppliedBuilder(BuildConfig buildConfig) =>
    buildConfig.builderDefinitions.values.any((def) => def.autoApply != AutoApply.none);
```

With those two, I think we could do something like this:

```suggestion
    if (buildConfigDefinesAutoAppliedBuilder(buildConfig) ||
        rootPackageReferencesDependencyInBuildYaml(package.name)) {
      packagesWithBuilders.add(package.name);
    }
```

And I guess maybe at that point it would be worth updating the list to be named `packagesWithConsumedBuilders` or something similar.

lib/dependency_validator.dart
```diff
@@ -51,41 +52,47 @@ Future<Null> run() async {
...
 
   logger.info('Validating dependencies for $packageName\n');
 
   // Find packages that provide executables.
   final packagesWithExecutables = Set<String>();
+  final packagesWithBuilders = Set<String>();
   final packageConfig = await findPackageConfig(Directory.current);
   for (final package in packageConfig.packages) {
+    // Search for builders
+    final buildConfig = await BuildConfig.fromBuildConfigDir(packageName, pubspec.dependencies.keys, package.root.path);
```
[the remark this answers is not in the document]
evanweible-wf: Also, since this process is more involved than the executable detection (have to read and parse two yaml files), I do think we should only do this for packages that are explicit dependencies and not every package in the package config file.

lib/dependency_validator.dart
evanweible-wf, line 78: Since we have the fully parsed builder definitions, we should be able to go a little further than just determining whether the dependency defines any builders. It's possible that the builder(s) defined in a package may not be auto-applied–in which case the package might actually be unused–but it's also possible that the root package explicitly enables a builder in its own build.yaml.

We can parse the BuildConfig for the root package prior to this loop:

```dart
  final rootBuildConfig = await BuildConfig.fromPackageDir('.');
```

Something like this should be able to determine whether the root package has a build.yaml config that references a builder from any given dependency:

```dart
  bool rootPackageReferencesDependencyInBuildYaml(String dependencyName) =>
    [
      ...rootBuildConfig.globalOptions.keys,
      for (final target in rootBuildConfig.buildTargets.values)
        ...target.builders.keys,
    ]
      .map(normalizeBuilderKeyUsage)
      .any((key) => key.startsWith('$dependencyName:');
```

And something like this should be able to determine whether a BuildConfig defines any builders that would be auto-applied to the root package:

```dart
  Future<bool> dependencyDefinesAutoAppliedBuilder(String path) async =>
    (await BuildConfig.fromPackageDir(path))
      .builderDefinitions
      .values
      .any((def) => def.autoApply != AutoApply.none);
```

With those two, I think we could do something like this:

```suggestion
    if (rootPackageReferencesDependencyInBuildYaml(package.name) ||
        await dependencyDefinesAutoAppliedBuilder(package.root.path)) {
      packagesWithBuilders.add(package.name);
    }
```

(Bonus: if the root package references a dependency, we get to skip parsing the BuildConfig for that dependency.)

And I guess maybe at that point it would be worth updating the list to be named `packagesWithConsumedBuilders` or something similar.

lib/dependency_validator.dart
```diff
@@ -266,15 +273,16 @@ Future<Null> run() async {
...
+  // Remove deps that provide executables, those are assumed to be used
+  logIntersection(Level.INFO, 'The following packages contain executables, they are assumed to be used:',
+      unusedDependencies, packagesWithExecutables);
   unusedDependencies.removeAll(packagesWithExecutables);
 
+  // Remove deps that provide builders, those are assumed to be used
+  logIntersection(Level.INFO, 'The following packages contain builders, they are assumed to be used:',
+      unusedDependencies, packagesWithBuilders);
+  unusedDependencies.removeAll(packagesWithBuilders);
+
```
evanweible-wf, line 285: With the above changes, we should be pretty confident that the `packagesWithBuilders` list includes packages that _are actually_ used, so I think we should reword it slightly and bump the level down so that it's only logged in verbose mode:

```suggestion
  // Remove deps that provide builders, those are assumed to be used
  logIntersection(Level.FINE, 'The following packages contain builders that are auto-applied or referenced in "build.yaml":',
      unusedDependencies, packagesWithBuilders);
  unusedDependencies.removeAll(packagesWithBuilders);

```

lib/dependency_validator.dart
```diff
@@ -317,22 +325,14 @@ Future<Null> run() async {
...
-    );
-  }
+  infractions.addAll(getDependenciesWithPins(pubspec.dependencies, ignoredPackages: ignoredPackages));
 
-  if (pubspecYaml.containsKey(devDependenciesKey)) {
-    infractions.addAll(
-      getDependenciesWithPins(pubspecYaml[devDependenciesKey], ignoredPackages: ignoredPackages),
-    );
-  }
+  infractions.addAll(getDependenciesWithPins(pubspec.devDependencies, ignoredPackages: ignoredPackages));
```
evanweible-wf, line 335: [nit]
```dart
final infractions = [
  ...getDependenciesWithPins(pubspec.dependencies, ignoredPackages: ignoredPackages),
  ...getDependenciesWithPins(pubspec.devDependencies, ignoredPackages: ignoredPackages),
];
```

lib/dependency_validator.dart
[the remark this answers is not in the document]
aaronstgeorge-wf: Oh oops I'm an idiot.

test/executable_test.dart
```diff
@@ -336,7 +348,7 @@ void main() {
       expect(result.stdout, contains('No fatal infractions found, valid is good to go!'));
     });
 
-    test('passes when there are unused known common binary packages', () async {
+    test('passes when dependencies not used in lib provide executables', () async {
```
evanweible-wf, line 351: I think this should be:
```suggestion
    test('passes when dependencies not used provide executables', () async {
```

test/executable_test.dart
```diff
@@ -345,13 +357,38 @@ void main() {
...
+        d.file('pubspec.yaml', pubspec),
+      ]).create();
+
+      final result = checkProject('${d.sandbox}/common_binaries');
+
+      expect(result.exitCode, 0);
+      expect(result.stdout, contains('No fatal infractions found, common_binaries is good to go!'));
+    });
+
+    test('passes when dependencies not used in lib provide auto applied builders', () async {
```
evanweible-wf, line 380:
```suggestion
    test('passes when dependencies are not imported but provide auto applied builders', () async {
```

test/executable_test.dart
```diff
@@ -369,6 +406,41 @@ void main() {
       expect(result.stdout, contains('No fatal infractions found, common_binaries is good to go!'));
     });
 
+    test('passes when dependencies not used in lib provide used builders', () async {
```
evanweible-wf, line 409:
```suggestion
    test('passes when dependencies are not imported but provide used builders', () async {
```

PR #107, opened 2024-03-11
Fail on non-dev packages with executables that are only used within `bin/`

### LINKED ISSUE #99: Old readme instructions
Readme stats to use the command 
`pub run dependency_validator` but it didn't work for me

https://github.com/Workiva/dependency_validator#usage

Found that 
`dart run dependency_validator`
is working

### LINKED ISSUE #106: Fail when a package is a non-dev dependency and is only referenced by its scripts
When a package isn't imported, dependency_validator will print the following message if it has bin scripts.

```The following packages contain executables, they are assumed to be used:```

However, a dependency that isn't referenced in lib should always be a dev_dependency. As a result, I propose that if a package is listed under `dependencies` and is unused, but has scripts, then dependency_validator should fail and suggest moving the package into `dev_dependencies`

### PR DESCRIPTION
**Pull request by matthewnitschke-wk**
## Motivation

- Closes #106
- Closes #99

## Changes
dependency_validator passes if there is a non-dev (`dependencies:`) dep, that contains executables, but is only referenced within the `bin/` directory

Under the following conditions, the dependency should be downgraded to a dev_dependency, and dependency_validator exits with a non-zero exit code

> [!IMPORTANT]
> Merging this PR has a chance to break existing implementations, we may want to do this as a major, depending on how best we want to support current implementations of dep validator

### CODE REVIEW

**Comment by aviary-wf** [comment]
## Security Insights

## Action Items
* Verify [`aviary.yaml`](https://wiki.atl.workiva.net/pages/viewpage.action?pageId=42762752#SecurityInsightFramework&Tooling-ConfiguringAviary) coverage of security relevant code
---

File: .github/workflows/dart_ci.yaml [hunk 1]
```diff
@@ -12,21 +12,17 @@ on:
...
+          sdk: 2.19.6
       - name: Install dependencies
         run: dart pub get
       - name: Validate dependencies
         run: dart run dependency_validator
-      - name: Check formatting
-        run: dart format --output=none --set-exit-if-changed .
-        if: ${{ matrix.sdk == 'stable' }}
+      # - name: Check formatting
+      #   run: dart format --output=none --set-exit-if-changed .
```
matthewnitschke-wk, line 25: Formatting hasn't been ran on the repo in awhile due to the `matrix.sdk == 'stable'` check

Will put up a different PR, running formatting, so as to not obscure the changes here with formatting diffs

dustyholmes-wf (approved):

evanweible-wf (dismissed):

matthewnitschke-wk:
This feature has been absorbed in the #111 branch

PR #138, opened 2024-12-23
Support Pub Workspaces

### LINKED ISSUE #137: Pub workspaces
The reliance on [`.dart_tool/package_config.json`](https://github.com/Workiva/dependency_validator/blob/16beb4df2be7a8d137a1b9a6a67d41781fd1deae/lib/src/dependency_validator.dart#L36) does not work with Dart's new [pub workspaces](https://dart.dev/tools/pub/workspaces) as those do not get generated in each individual package in a workspace. In workspaces, only a root configuration is generated.

Levi-Lesches:
## Motivation
  

Fixes #137 by supporting use in a mono-repo

## Changes
  

In a regular package: nothing
In a sub-package of a workspace: nothing
In a workspace: instead of failing, recursively run itself on all sub-packages

I did remove the error condition that `dart pub get` must be run before running the tool, since that was throwing errors for workspaces, and I did not see a use for it (the `pubspec.lock` or `.dart_tool` files are never checked, only the code and `pubspec.yaml`).

## Testing/QA Instructions
  

Looking into adding unit tests now, but basically, [create a workspace](https://dart.dev/tools/pub/workspaces) and try running the tool in there. 

For now, I had to add a dependency override on pubspec_parse until https://github.com/dart-lang/tools/pull/1948 is merged. For some reason that's affecting the tests, so many of them are failing as they can't override the package while executing. 

@evanweible-wf for review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 5.0.0
+
+- **Breaking change**: Requires Dart 3.0 or above
+- Added support for Pub Workspaces (monorepos)
+
 # 4.1.1
 
 - Update the output of parse failures to include the path to the file which failed to parse
--- a/lib/src/dependency_validator.dart
+++ b/lib/src/dependency_validator.dart
@@ -32,6 +32,7 @@ Future<bool> checkPackage({required String root}) async {
   var result = true;
   if (!File('$root/pubspec.yaml').existsSync()) {
     logger.shout(red.wrap('pubspec.yaml not found'));
+    logger.fine('Path: $root/pubspec.yaml');
     return false;
   }
 
@@ -55,7 +56,7 @@ Future<bool> checkPackage({required String root}) async {
       .map((s) {
         try {
           // Globs don't support './' paths. Use posix to avoid '\' paths
-          final globPath = p.posix.normalize('$root/$s');
+          final globPath = '$root/$s'.replaceAll(r'\', '/');
           return Glob(globPath);
         } catch (_, __) {
           logger.shout(yellow.wrap('invalid glob syntax: "$s"'));
@@ -82,7 +83,7 @@ Future<bool> checkPackage({required String root}) async {
   if (pubspec.isWorkspaceRoot) {
     logger.fine('In a workspace. Recursing through sub-packages...');
     for (final package in pubspec.workspace ?? []) {
-      subResult &= await checkPackage(root: package);
+      subResult &= await checkPackage(root: '$root/$package');
       logger.info('');
     }
   }
@@ -142,10 +143,17 @@ Future<bool> checkPackage({required String root}) async {
   logger.fine('packages used in public facing files:\n'
       '${bulletItems(packagesUsedInPublicFiles)}\n');
 
-  final publicDirGlobs = [for (final dir in publicDirs) Glob('$dir**')];
-  final subpackageGlobs = [
-    for (final subpackage in pubspec.workspace ?? []) Glob('$subpackage**')
+  final publicDirGlobs = [
+    for (final dir in publicDirs)
+      Glob('$dir**'.replaceAll(r'\', '/')),
   ];
+
+  final subpackageGlobs = [
+    for (final subpackage in pubspec.workspace ?? [])
+      Glob('$root/$subpackage**'.replaceAll(r'\', '/')),
+  ];
+
+  logger.fine('subpackage globs: $subpackageGlobs');
 
   final nonPublicDartFiles = listDartFilesIn(
     '$root/',
@@ -303,9 +311,11 @@ Future<bool> checkPackage({required String root}) async {
           .any((key) => key.startsWith('$dependencyName:'));
 
   final packagesWithConsumedBuilders = Set<String>();
-  for (final package in unusedDependencies.map((name) => packageConfig[name])) {
+  for (final name in unusedDependencies) {
+    final package = packageConfig[name];
+    if (package == null) continue;
     // Check if a builder is used from this package
-    if (rootPackageReferencesDependencyInBuildYaml(package!.name) ||
+    if (rootPackageReferencesDependencyInBuildYaml(package.name) ||
         await dependencyDefinesAutoAppliedBuilder(p.fromUri(package.root))) {
       packagesWithConsumedBuilders.add(package.name);
     }
@@ -320,14 +330,21 @@ Future<bool> checkPackage({required String root}) async {
   unusedDependencies.removeAll(packagesWithConsumedBuilders);
 
   // Remove deps that provide executables, those are assumed to be used
-  final packagesWithExecutables = unusedDependencies.where((name) {
+  bool providesExecutable(String name) {
     final package = packageConfig[name];
-    final binDir = Directory(p.join(p.fromUri(package!.root), 'bin'));
+    if (package == null) return false;
+    final binDir = Directory(p.join(p.fromUri(package.root), 'bin'));
     if (!binDir.existsSync()) return false;
 
     // Search for executables, if found we assume they are used
     return binDir.listSync().any((entity) => entity.path.endsWith('.dart'));
-  }).toSet();
+  }
+
+  final packagesWithExecutables = {
+    for (final package in unusedDependencies)
+      if (providesExecutable(package))
+        package,
+  };
 
   final nonDevPackagesWithExecutables =
       packagesWithExecutables.where(pubspec.dependencies.containsKey).toSet();
--- a/lib/src/dependency_validator.dart
+++ b/lib/src/dependency_validator.dart
@@ -73,8 +73,10 @@ Future<bool> checkPackage({required String root}) async {
 
   // Read and parse the pubspec.yaml in the current working directory.
   final pubspecFile = File('$root/pubspec.yaml');
-  final pubspec =
-      Pubspec.parse(pubspecFile.readAsStringSync(), sourceUrl: pubspecFile.uri,);
+  final pubspec = Pubspec.parse(
+    pubspecFile.readAsStringSync(),
+    sourceUrl: pubspecFile.uri,
+  );
 
   var subResult = true;
   if (pubspec.isWorkspaceRoot) {
@@ -141,14 +143,22 @@ Future<bool> checkPackage({required String root}) async {
       '${bulletItems(packagesUsedInPublicFiles)}\n');
 
   final publicDirGlobs = [for (final dir in publicDirs) Glob('$dir**')];
-  final subpackageGlobs = [for (final subpackage in pubspec.workspace ?? []) Glob('$subpackage**')];
+  final subpackageGlobs = [
+    for (final subpackage in pubspec.workspace ?? []) Glob('$subpackage**')
+  ];
 
-  final nonPublicDartFiles =
-      listDartFilesIn('$root/', [...excludes, ...publicDirGlobs, ...subpackageGlobs]);
-  final nonPublicScssFiles =
-      listScssFilesIn('$root/', [...excludes, ...publicDirGlobs, ...subpackageGlobs]);
-  final nonPublicLessFiles =
-      listLessFilesIn('$root/', [...excludes, ...publicDirGlobs, ...subpackageGlobs]);
+  final nonPublicDartFiles = listDartFilesIn(
+    '$root/',
+    [...excludes, ...publicDirGlobs, ...subpackageGlobs],
+  );
+  final nonPublicScssFiles = listScssFilesIn(
+    '$root/',
+    [...excludes, ...publicDirGlobs, ...subpackageGlobs],
+  );
+  final nonPublicLessFiles = listLessFilesIn(
+    '$root/',
+    [...excludes, ...publicDirGlobs, ...subpackageGlobs],
+  );
 
   logger
     ..fine('non-public dart files:\n'
@@ -302,10 +312,11 @@ Future<bool> checkPackage({required String root}) async {
   }
 
   logIntersection(
-      Level.FINE,
-      'The following packages contain builders that are auto-applied or referenced in "build.yaml"',
-      unusedDependencies,
-      packagesWithConsumedBuilders,);
+    Level.FINE,
+    'The following packages contain builders that are auto-applied or referenced in "build.yaml"',
+    unusedDependencies,
+    packagesWithConsumedBuilders,
+  );
   unusedDependencies.removeAll(packagesWithConsumedBuilders);
 
   // Remove deps that provide executables, those are assumed to be used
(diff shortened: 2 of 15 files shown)


matthewnitschke-wk:
Thanks for the contribution! I'll take a look in a bit!

matthewnitschke-wk:
@Levi-Lesches 

Approach looks good to me! but I do have a few questions about some edge cases of the pub workspace implementation

Its not [documented super well](https://dart.dev/tools/pub/workspaces), but it does appear that you are still able to implement dart source in the root (interestingly any dependencies added to the root are available to all sub-packages, but this messes up publishing of each sub-package, which would to be a negative, and transitively a fantastic opportunity for dependency_validator to catch)

Because of the above, I think it would be best to not stop evaluation of the root `lib/` assets if `workspaces` is declared, like is currently implemented

Additionally, I noticed that when one of the sub-packages has a violation, other subpackages are not outputted as a "success", the output is empty. Visually, it would be nice to see each subpackage outputted, and whether it passed validation

Finally, dart apis [support a `-C` option](https://dart.dev/tools/pub/workspaces#running-a-command-in-a-specific-workspace-package) to execute on provided subpackages. Not something that I'd hold this PR back on, but it might be nice to mirror this option and allow dependency_validator to be ran from the root, but only evaluate a single sub-package

Since we're currently blocked on https://github.com/dart-lang/tools/pull/1948, I'll go ahead and convert this PR to a draft, and apply a Hold label. Feel free to ping me directly when dependent change is merged to the upstream, and the above issues resolved

lib/src/dependency_validator.dart
```diff
@@ -28,31 +28,28 @@ import 'pubspec_config.dart';
...
+    );
     if (pubspecConfig.isNotEmpty) {
       logger.warning(yellow.wrap(
           'Configuring dependency_validator in pubspec.yaml is deprecated.\n'
           'Use dart_dependency_validator.yaml instead.'));
     }
     config = pubspecConfig.dependencyValidator;
   }
+
   final excludes = config.exclude
```
matthewnitschke-wk, line 53: Realized this a bit after the fact, but since we're re-using the same config values as the root, these `exclude` paths will be re-evaluated on subpackages

Meaning, if we exclude:

```yaml
exclude:
  - lib/src/some_path/**
```

This would exclude `lib/src/some_path` in the root, and also exclude that same path in each subpackage

I think a better implementation for this is to treat each path as if it came from the root, which should be pretty easy to achieve using [`relative`](https://pub.dev/documentation/path/latest/path/relative.html) in conjunction with the `root` parameter

gabrielgarciagava:
@matthewnitschke-wk Just a heads up regarding:

> Since we're currently blocked on https://github.com/dart-lang/tools/pull/1948, I'll go ahead and convert this PR to a draft, and apply a Hold label. Feel free to ping me directly when dependent change is merged to the upstream, and the above issues resolved

The linked PR is merged now.

matthewnitschke-wk:
> @matthewnitschke-wk Just a heads up regarding:
> 
> 
> 
> > Since we're currently blocked on https://github.com/dart-lang/tools/pull/1948, I'll go ahead and convert this PR to a draft, and apply a Hold label. Feel free to ping me directly when dependent change is merged to the upstream, and the above issues resolved
> 
> 
> 
> The linked PR is merged now.

Thanks! Looks like we still need the package published: https://pub.dev/packages/pubspec_parse/changelog

lib/src/dependency_validator.dart
Levi-Lesches: Hm, I think you may be mistaken. The workspace logic just calls the whole function recursively, so the config is indeed re-loaded per subpackage. For example: 

```text
root/
├─ lib/
│  ├─ main.dart
├─ pkg1/
│  ├─ lib/
│  │  ├─ main.dart
├─ pkg2/
│  ├─ lib/
│  │  ├─ main.dart
```

Say that `root` and `pkg1` both import Flutter, but don't declare it in their pubspec. And the user has the following config in the root, with no config in any other dir: 

```yaml
exclude:
  - 'lib/main.dart'
```

Then I get the following output: 
```console
Validating dependencies for pkg1...
These packages are used in lib/ but are not dependencies:
  * flutter

Validating dependencies for pkg2...
✓ No dependency issues found!

Validating dependencies for temp...
✓ No dependency issues found!
```

Specifically, I have this line when loading the config: 

```dart
  final configFile = File('$root/dart_dependency_validator.yaml');
```

which I believe is what you were talking about.

Levi-Lesches:
> Because of the above, I think it would be best to not stop evaluation of the root lib/ assets if workspaces is declared, like is currently implemented

> Additionally, I noticed that when one of the sub-packages has a violation, other subpackages are not outputted as a "success", the output is empty. Visually, it would be nice to see each subpackage outputted, and whether it passed validation

Fixed both

```console
Validating dependencies for pkg1...
✓ No dependency issues found!

Validating dependencies for pkg2...
These packages may be unused, or you may be using assets from these packages:
  * meta

Validating dependencies for root...
These packages are used in lib/ but are not dependencies:
  * flutter
```

>  (interestingly any dependencies added to the root are available to all sub-packages, but this messes up publishing of each sub-package, which would to be a negative, and transitively a fantastic opportunity for dependency_validator to catch)

Already handled! Say `root` imports and declares `package:http`, and `pkg1` uses `http` without declaring it:

```console
Validating dependencies for pkg1...
These packages are used in lib/ but are not dependencies:
  * http

Validating dependencies for pkg2...
✓ No dependency issues found!

Validating dependencies for root...
✓ No dependency issues found!
```

Worth noting that `dart analyze` already flags this as [`depend_on_referenced_packages`](https://dart.dev/diagnostics/depend_on_referenced_packages)
 
> Finally, dart apis [support a -C option](https://dart.dev/tools/pub/workspaces#running-a-command-in-a-specific-workspace-package) to execute on provided subpackages. Not something that I'd hold this PR back on, but it might be nice to mirror this option and allow dependency_validator to be ran from the root, but only evaluate a single sub-package

Done!

```console
$ dart run dependency_validator --directory pkg1
$ dart run dependency_validator -C pkg1

Building package executable... 
Built dependency_validator:dependency_validator.

Validating dependencies for pkg1...
✓ No dependency issues found!
```

> Thanks! Looks like we still need the package published

Version 1.5.0 will probably hit soon, I know they're working on some other small PRs first (glad I managed to get mine in!)

- https://github.com/dart-lang/tools/pull/1952
- https://github.com/dart-lang/tools/pull/1951
- https://github.com/dart-lang/tools/pull/1946

----

Some tests are still failing so I will take a look again tomorrow

lib/src/dependency_validator.dart
matthewnitschke-wk: Yep, it was thinking that all paths should be relative to the root, and not re-evaluated per package

where in your above example, if we wanted to exclude `pkg1` and `root`, we'd have to specify the following:

```yaml
# Exclusion paths are always evaluated relative to the root of the project
exclude:
  - 'lib/main.dart'
  - 'pkg1/lib/main.dart'
```

I think this approach is clearer than applying the same exclude paths per subpackage, and allows for more flexibility for the cases where not all subpackages might not follow the same folder structure

I'm happy to discuss this further if you disagree though

matthewbelisle-wf:
Going to close and re-open this, trying to fix one of the CI errors.

lib/src/dependency_validator.dart
Levi-Lesches:
```yaml
exclude:
  - 'lib/main.dart'
  - 'pkg1/lib/main.dart'
```

Just to clarify, that wouldn't work the PR right now. You'd have to provide a config with `lib/main.dart` for `root` and `pkg1`, so two configs that look like this: 

```yaml
exclude:
  - 'lib/main.dart'
```

> I think this approach is clearer than applying the same exclude paths per subpackage, and allows for more flexibility for the cases where not all subpackages might not follow the same folder structure

I think being able to provide a new config per package is actually _more_ flexible, if not a bit more verbose, because it keeps the packages completely separate and independent from each other. It's definitely _easier_ to just provide a l ist of deep paths in the root, and only have one config file, but that also makes the behavior less clear: what happens if the root config mentions paths in the sub-packages, but the sub-packages provide their own configs? 

> Yep, it was thinking that all paths should be relative to the root, and not re-evaluated per package

I think I get what you mean here though: I just tried it, and having a config file in `pkg1` that says to exclude `lib/main.dart` still results in an error for `pkg1`, because the tool is ignoring `root/lib/main.dart` and not `root/pkg1/lib/main.dart`. I just fixed that, so now, excluded paths apply to the same package as their config file.

matthewnitschke-wk: > because it keeps the packages completely separate and independent from each other

Great point, I was slightly mistaken a bit on the intention of multiple configs. I like that design and approach

> I think I get what you mean here though: I just tried it,...

Perfect, "excluded paths apply to the same package as their config file" I think this is the best approach

Levi-Lesches:
I'm wondering if I can make testing easier by basically wrapping all existing tests, and then running them twice: once as normal packages and one as a sub-package in a workspace. How feasible do you think that is?

pubspec.yaml
```diff
@@ -4,7 +4,7 @@ description: Checks for missing, under-promoted, over-promoted, and unused depen
 homepage: https://github.com/Workiva/dependency_validator
 
 environment:
-  sdk: '>=2.19.0 <3.0.0'
+  sdk: '^3.0.0'
```
matthewnitschke-wk, line 7: Could we update this to still allow for dart v2? We still have a few consumers of dependency_validator which have yet to migrate

```suggestion
  sdk: '>=2.19.0 <4.0.0'
```

matthewnitschke-wk:
> I'm wondering if I can make testing easier by basically wrapping all existing tests, and then running them twice: once as normal packages and one as a sub-package in a workspace. How feasible do you think that is?

My personal preference would be to write specific tests for the new functionality that workspace support implements:
- execution of subpackages
  - with respecting nested `dart_dependency_validator.yaml` files
- path normalization
- evaluation of the root package

That way we can isolate what specifically needs to be tested for workspaces, and what is more generally applicable to individual package evaluation

pubspec.yaml
Levi-Lesches: That's due to pubspec_parse, which, since the 1.0 that you were using before, has required more and more recent versions of Dart. In fact, version 1.5 [will require Dart 3.6](https://github.com/dart-lang/tools/pull/1946). 

This shouldn't be an issue though, since Dart [treats versions after 2.12 to be compatible with 3.0](https://dart.dev/resources/dart-3-migration#dart-3-backwards-compatibility). The only things that would cause a breakage from 2.19 are (list in the link above as "unversioned changes"): 
- packages that don't have null safety, or use dependencies that don't have null safety. Given that null safety came out 4 years ago and almost all packages one would depend on have migrated as well, I'd think it makes sense to no longer support pre-null-safe code in future updates, and if customers still don't have it, they can continue to use the current version of dependency_validator. In other words, customers without sound null safety are missing out on _many_ more important features than just this tool.
- Some APIs that had already been deprecated for years were removed
- Some tools were renamed (eg, `stagehand` -> `dart create`)
- Some minor other changes, see the link above

Levi-Lesches:
@matthewnitschke-wk pubspec_parse v1.5 has been published, and I added 11 new tests. 

I also added a `.toJson()` for pubspecs and configs, which really helped when writing clean tests. It may be possible to refactor the existing tests to take advantage of this like the workspace tests too.

Levi-Lesches:
Okay I went ahead and gave the same treatment to all the other tests in `executable_tests.dart`: 

Before: 
```console
$ dart test test\executable_test.dart
02:41 +25: All tests passed!
```

After: 
```console
$ dart test test\executable_test.dart
01:50 +25: All tests passed!
```

This commit also fixes a bug and leverages that `pubspec_parse` requires Dart 3 (see above comment) to remove all those dependency overrides in the tests. But even so, this new commit means _you don't have to hand-write pubspecs anymore!_ All in all, the net diff of that commit is -237 lines and no more hand-written yaml.

matthewnitschke-wk:
@Levi-Lesches Excellent updates! Thanks for optimizing our tests!!

Now that we're updating the minimimum dart version to 3, we'll need to update `ci.yaml` to:
```yaml
name: CI

on:
  pull_request:
  push:
    branches:
      - 'master'

jobs:
  build:
    uses: Workiva/gha-dart-oss/.github/workflows/build.yaml@v0.1.7
    with:
      sdk: stable

  checks:
    uses: Workiva/gha-dart-oss/.github/workflows/checks.yaml@v0.1.7
    with:
      sdk: stable

  unit-tests:
    uses: Workiva/gha-dart-oss/.github/workflows/test-unit.yaml@v0.1.7
    with:
      sdk: stable
```

and

```yaml
name: Publish

on:
  push:
    tags:
      - '[0-9]+.[0-9]+.[0-9]+'

permissions:
  contents: write
  id-token: write
  pull-requests: write

jobs:
  publish:
    uses: Workiva/gha-dart-oss/.github/workflows/publish.yaml@v0.1.7
    with:
      sdk: stable
```

Once that's updated, we should be able to get passing CI, and merge this work!

As a side note, I've annotated this PR with `semver: Major`, meaning the next tagged release should be: `v5.0.0`. This is mostly due to the supported dart sdk version increasing, which is a breaking change

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 5.0.0
+
+- **Breaking change**: Requires Dart 3.0 or above
```
dustinlessard-wf, line 3: Is Dart 3 strictly required?  This is a problem for us as we haven't yet completed the migration to Dart 3 and are still on Dart 2.19.

dustinlessard-wf (approved):

dustinlessard-wf:
@Levi-Lesches , we're intending on merging and releasing this change today. 🙇🏼

dustinlessard-wf:
Closing and reopening the pr as it looks like the unit-tests workflow is stuck.

dustinlessard-wf:
There's one issue that we need to resolve first.  It looks like the unit-tests aren't actually running.
Screenshot 2025-01-10 at 7 40 45 AM
I was hoping a close/reopen would be a quick-fix, but it looks like something else is going on.

Levi-Lesches:
@dustinlessard-wf I looked through your ci for a bit, and I found this: 

chrome-platform: false
dart-dev: false
build-runner: true
build-test: false

The tests are only configured to run if: 
- both builds are false, or 
- both builds are true

Since this case is neither, both workflows are being skipped (see: gha-oss/test-unit.yaml)

matthewnitschke-wk:
@Levi-Lesches This was a nuance of how we handle "required" workflows at Workiva. Fixed it from an external system we have. Fixed that issue and it should be good to go now!

Just running through some manual testing and then we should be good to merge!

matthewnitschke-wk:
🚀 @Workiva/release-management-p 🚢

matthewnitschke-wk:
semver +1

Increases the accepted dart version from 2 to 3

rmconsole-wf (approved):
+1 from RM

Levi-Lesches:
Thanks for the quick merge and publish! 

I love this tool and use it for everything, so I'm very happy to get the chance to contribute

PR #148, opened 2025-03-06
FEDX-2722: Update the `ignore` field to ignore everything, not just specific checks

matthewnitschke-wk:
# [FEDX-2722](https://jira.atl.workiva.net/browse/FEDX-2722)
Issue Status

## Motivation

Currently, the `ignore` field only applies to some rules in the dependency_validator execution process

If you're explicitly ignoring a dependency, it doesn't make sense that _some_ rules can still fail. As such, this PR changes the expectation of evaluation, where when `ignore` exists for a package, its full ignored from dep evaluation

> [!TIP]
> Note, this PR's base is `v4` and not `master`
>
> We need to backpatch the v4 version with this change, and this will be manually released once merged into v4. We will create a similar PR in master so the functionality between these branches doesn't change  

## Testing/QA Instructions
- [ ] Consume in a package that fails the "The following packages contain executables, and are only used outside of lib/. These should be downgraded to dev_dependencies:" rule, and ensure that adding `ignore` indeed ignores it

### DIFF
--- a/lib/src/dependency_validator.dart
+++ b/lib/src/dependency_validator.dart
@@ -310,18 +310,6 @@ Future<void> run() async {
     return binDir.listSync().any((entity) => entity.path.endsWith('.dart'));
   }).toSet();
 
-  final nonDevPackagesWithExecutables =
-      packagesWithExecutables.where(pubspec.dependencies.containsKey).toSet();
-  if (nonDevPackagesWithExecutables.isNotEmpty) {
-    logIntersection(
-      Level.WARNING,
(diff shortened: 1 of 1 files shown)


nicbiggs-wk (approved):

danielszopa-wk (approved):
+1 I can confirm this fixes the issue [this xbrl-module PR](https://github.com/Workiva/xbrl-module/pull/7426) was hitting.

brianneal-wf (approved):

alanknight-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #158, opened 2025-10-06
Support YAML list format for analysis_options includes, and update tests accordingly

leventkantaroglu:
The `getAnalysisOptionsIncludePackage` method previously only supported single string includes in _analysis_options.yaml_ files, but modern Dart/Flutter projects often use YAML list format for includes to reference multiple configuration packages. This enhancement adds support for the list format commonly used in projects like:

```
include:
  - package:flutter_lints/flutter.yaml
  - package:pedantic/analysis_options.1.8.0.yaml
 ```

This change improves the dependency validator's ability to detect analysis option packages that should be included in dependency validation, regardless of the YAML format used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 5.0.3
+
+- Add YAML list format support for analysis_options file
+
 # 5.0.2
 
 - Fixed changelog entries caused from mis-configured auto-release

### CODE REVIEW

File: lib/src/dependency_validator.dart [hunk 1]
```diff
@@ -199,7 +199,8 @@ Future<bool> checkPackage({required String root}) async {
   final packagesUsedOutsidePublicDirs = <String>{
     // For more info on analysis options:
     // https://dart.dev/guides/language/analysis-options#the-analysis-options-file
-    if (optionsIncludePackage != null) optionsIncludePackage,
+    if (optionsIncludePackage != null && optionsIncludePackage.isNotEmpty)
```
alanknight-wk, line 202: nit: why is this check necessary? Wouldn't `...[]` do nothing?

lib/src/utils.dart
```diff
@@ -31,19 +31,45 @@ final Logger logger = Logger('dependency_validator');
...
 
   final YamlMap? analysisOptions = loadYaml(optionsFile.readAsStringSync());
   if (analysisOptions == null) return null;
 
-  final String? include = analysisOptions['include'];
-  if (include == null || !include.startsWith('package:')) return null;
+  final dynamic include = analysisOptions['include'];
+  if (include == null) return null;
+
+  final Set<String> packageNames = <String>{};
```
alanknight-wk, line 54: nit: We generally follow this rule https://dart.dev/effective-dart/design#dont-redundantly-type-annotate-initialized-local-variables so the explicit declaration isn't necessary.

lib/src/utils.dart
```diff
@@ -31,19 +31,45 @@ final Logger logger = Logger('dependency_validator');
...
+Set<String>? getAnalysisOptionsIncludePackage({String? path}) {
   final optionsFile = File(p.join(path ?? p.current, 'analysis_options.yaml'));
   if (!optionsFile.existsSync()) return null;
 
   final YamlMap? analysisOptions = loadYaml(optionsFile.readAsStringSync());
   if (analysisOptions == null) return null;
 
-  final String? include = analysisOptions['include'];
-  if (include == null || !include.startsWith('package:')) return null;
+  final dynamic include = analysisOptions['include'];
```
alanknight-wk, line 51: nit: Slight preference for `Object?` over `dynamic`.

alanknight-wk (commented):
Thanks for the contribution!
A couple of very small style nits.
The checker also wants the files formatted, and an entry in the CHANGELOG.

kongo2002:
Hi, what's the status on this issue? We recently ran into the same problem

Issue #176, opened 2026-02-26
Glob workspaces support


jkoenig134:
As flutter workspaces now [supports glob](https://dart.dev/tools/pub/workspaces#glob-pattern-support) it would be very nice if dependency_validator also supports it.

PR #190, opened 2026-09-08
FEDX-7266: Support glob patterns in dependency_validator workspace resolution

engops-wk:
Opened by @matthewnitschke-wk with the ai-sdlc workflow for FEDX-7266

## Motivation
Dart/Flutter pub workspaces support glob patterns (e.g. `packages/*`) for declaring workspace members, but `dependency_validator` only handled literal paths in the `workspace` field of the pubspec. As a result, workspace configurations using glob patterns did not resolve correctly — sub-packages could be skipped during recursion, and the glob used to exclude sub-package files from the root package's unused-dependency check was built incorrectly. This change adds glob support so workspace resolution matches `dart pub`'s behavior (ref: Workiva/dependency_validator#176).

## Changes
- Added `resolveWorkspaceMembers` and `looksLikeGlob` in `lib/src/utils.dart`. Glob patterns are expanded via `package:glob` to concrete directories under the workspace root that contain a `pubspec.yaml`; literal paths are returned unchanged; invalid glob syntax is caught and skipped.
- Updated `checkPackage` in `lib/src/dependency_validator.dart` to resolve workspace members once (via `resolveWorkspaceMembers`) and reuse the resolved list both for recursing into sub-packages and for constructing the `subpackageGlobs` used to exclude sub-package files from the top-level dependency scan. Path joining now uses `p.join`.
- Added a changelog entry describing the new glob support.
- Extended the `checkWorkspace` test helper (`test/utils.dart`) with `workspaceMembers`, `subpackagePath`, and `extraDirs` options to support testing glob-based workspace layouts.
- Added unit tests for `resolveWorkspaceMembers`/`looksLikeGlob` (`test/utils_test.dart`) and integration tests for glob-based workspace resolution (`test/workspace_test.dart`), covering literal paths, glob expansion, directories without `pubspec.yaml`, and propagation of validation failures from glob-matched sub-packages.

## Testing/QA Instructions
1. Run `dart test` and confirm all tests pass, including the new tests covering glob resolution.
2. Run `dart test test/utils_test.dart --name resolveWorkspaceMembers` to verify glob expansion logic in isolation (literal paths pass through, `packages/*` expands to matching sub-packages with a `pubspec.yaml`, mixed literal/glob patterns resolve correctly).
3. Run `dart test test/workspace_test.dart --name "glob workspace patterns"` to verify end-to-end behavior: a workspace declaring `packages/*` recurses into matching sub-packages, directories without `pubspec.yaml` are ignored, and unused-dependency issues inside a glob-matched sub-package still fail the overall check.
4. CI passing on the above automated tests is sufficient; no additional manual QA steps are required beyond optionally exercising `dependency_validator` against a local workspace using a glob pattern in its `workspace` field.

- [x] This change is covered by sufficient tests (existing or new), OR I have provided a reason for why tests are not necessary
- [x] I have updated the CHANGELOG.md, OR I have provided a reason for why this change doesn't need it

## Intent

Add glob pattern support to workspace member resolution in dependency_validator. Currently, the tool only handles literal paths in workspace configurations, causing issues when workspaces declare members using glob patterns like `packages/*`. This change adds glob expansion via the `package:glob` library, aligning dependency_validator's workspace resolution behavior with `dart pub` and fixing sub-package discovery and unused-dependency checks for glob-based workspace layouts.

## How To QA

1. 1. Run full test suite with `dart test` and confirm all tests pass, including new/updated tests in test/utils_test.dart and test/workspace_test.dart
2. 2. Run `dart test test/utils_test.dart --name resolveWorkspaceMembers` and verify: literal paths pass through unchanged, `packages/*` expands only to directories with pubspec.yaml, and mixed literal/glob patterns resolve correctly
3. 3. Run `dart test test/workspace_test.dart --name "glob workspace patterns"` and confirm: `packages/*` correctly recurses into matching sub-packages, directories without pubspec.yaml are ignored, and unused-dependency issues in glob-matched sub-packages cause overall check to fail
4. 4. Optionally, create a local package with pubspec_workspace.yaml using `workspace: ['packages/*']` and manually verify dependency_validator correctly recurses into glob-matched sub-packages and excludes their files from root package's unused-dependency scan

### DIFF
--- a/lib/src/dependency_validator.dart
+++ b/lib/src/dependency_validator.dart
@@ -80,10 +80,16 @@ Future<bool> checkPackage({required String root}) async {
   );
 
   var subResult = true;
+  final workspaceMembers = pubspec.isWorkspaceRoot
+      ? resolveWorkspaceMembers(root, pubspec.workspace ?? [])
+      : const <String>[];
   if (pubspec.isWorkspaceRoot) {
     logger.fine('In a workspace. Recursing through sub-packages...');
-    for (final package in pubspec.workspace ?? []) {
-      subResult &= await checkPackage(root: '$root/$package');
+    logger.fine(
+      'workspace members:\n${bulletItems(workspaceMembers)}\n',
+    );
+    for (final package in workspaceMembers) {
+      subResult &= await checkPackage(root: p.join(root, package));
       logger.info('');
     }
   }
@@ -157,8 +163,8 @@ Future<bool> checkPackage({required String root}) async {
   final publicDirGlobs = [for (final dir in publicDirs) makeGlob('$dir**')];
 
   final subpackageGlobs = [
-    for (final subpackage in pubspec.workspace ?? [])
-      makeGlob('$root/$subpackage**'),
+    for (final subpackage in workspaceMembers)
+      makeGlob('${p.join(root, subpackage)}**'),
   ];
 
   logger.fine('subpackage globs: $subpackageGlobs');
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -15,6 +15,7 @@
 import 'dart:io';
 
 import 'package:glob/glob.dart';
+import 'package:glob/list_local_fs.dart';
 import 'package:io/ansi.dart';
 import 'package:pub_semver/pub_semver.dart';
 import 'package:logging/logging.dart';
@@ -209,3 +210,39 @@ extension PubspecUtils on Pubspec {
 /// This function removes `./` paths and replaces all `\` with `/`.
 Glob makeGlob(String path) =>
     Glob(p.posix.normalize(path.replaceAll(r'\', '/')));
+
+/// Returns whether [path] looks like a glob pattern.
+bool looksLikeGlob(String path) => Glob.quote(path) != path;
+
+/// Resolves workspace member patterns to concrete package paths.
+///
+/// Glob patterns are expanded to directories beneath [root] that contain a
+/// `pubspec.yaml`, matching how `dart pub` resolves workspace members.
+/// Non-glob entries are returned as-is.
+List<String> resolveWorkspaceMembers(
+  String root,
+  Iterable<String> patterns,
+) {
+  final members = <String>{};
+  for (final pattern in patterns) {
+    final normalized = p.posix.normalize(pattern.replaceAll(r'\', '/'));
+    if (looksLikeGlob(normalized)) {
+      try {
+        final glob = makeGlob(normalized);
+        for (final entity in glob.listSync(root: root)) {
+          if (entity is! Directory) continue;
+          final pubspec = File(p.join(entity.path, 'pubspec.yaml'));
+          if (!pubspec.existsSync()) continue;
+          members.add(
+            p.posix.normalize(p.relative(entity.path, from: root)),
+          );
+        }
+      } on FormatException {
+        // Invalid glob syntax; skip this pattern.
+      }
+    } else {
+      members.add(normalized);
+    }
+  }
+  return members.toList()..sort();
+}
--- a/test/utils.dart
+++ b/test/utils.dart
@@ -74,6 +74,9 @@ Future<void> checkWorkspace({
   required Map<String, Dependency> subpackageDeps,
   required List<d.Descriptor> workspace,
   required List<d.Descriptor> subpackage,
+  List<String> workspaceMembers = const ['subpackage'],
+  String subpackagePath = 'subpackage',
+  List<d.Descriptor> extraDirs = const [],
   DepValidatorConfig? workspaceConfig,
   DepValidatorConfig? subpackageConfig,
   Level logLevel = Level.OFF,
@@ -83,7 +86,7 @@ Future<void> checkWorkspace({
     'workspace',
     environment: requireDart36,
     dependencies: workspaceDeps,
-    workspace: ['subpackage'],
+    workspace: workspaceMembers,
   );
   final subpackagePubspec = Pubspec(
     'subpackage',
@@ -99,7 +102,8 @@ Future<void> checkWorkspace({
         'dart_dependency_validator.yaml',
         jsonEncode(workspaceConfig.toJson()),
       ),
-    d.dir('subpackage', [
+    ...extraDirs,
+    d.dir(subpackagePath, [
       ...subpackage,
       d.file('pubspec.yaml', jsonEncode(subpackagePubspec.toJson())),
       if (subpackageConfig != null)
--- a/test/workspace_test.dart
+++ b/test/workspace_test.dart
@@ -149,4 +149,44 @@ void main() => group('Workspaces', () {
           ),
         );
       });
+
+      group('glob workspace patterns', () {
+        test(
+          'resolves packages/* to subpackages with pubspec.yaml',
+          () => checkWorkspace(
+            workspace: [],
+            workspaceDeps: {},
+            workspaceMembers: ['packages/*'],
+            subpackagePath: 'packages/subpackage',
+            subpackage: usesHttp,
+            subpackageDeps: dependsOnHttp,
+          ),
+        );
+
+        test(
+          'ignores directories without pubspec.yaml',
+          () => checkWorkspace(
+            workspace: [],
+            workspaceDeps: {},
+            workspaceMembers: ['packages/*'],
+            subpackagePath: 'packages/subpackage',
+            extraDirs: [d.dir('packages/no_pubspec', [])],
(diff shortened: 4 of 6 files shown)


lib/src/utils.dart
```diff
@@ -209,3 +210,39 @@ extension PubspecUtils on Pubspec {
...
+        final glob = makeGlob(normalized);
+        for (final entity in glob.listSync(root: root)) {
+          if (entity is! Directory) continue;
+          final pubspec = File(p.join(entity.path, 'pubspec.yaml'));
+          if (!pubspec.existsSync()) continue;
+          members.add(
+            p.posix.normalize(p.relative(entity.path, from: root)),
+          );
+        }
+      } on FormatException {
```
engops-wk, line 240: Silently swallowing invalid glob syntax (and silently returning nothing when a pattern matches zero packages) means a typo'd `workspace:` entry makes validation pass while checking nothing — a regression from today's loud `pubspec.yaml not found` failure for literal paths. Please log both cases, mirroring `logger.shout(yellow.wrap('invalid glob syntax: "$s"'))` in `checkPackage`.

lib/src/utils.dart
```diff
@@ -209,3 +210,39 @@ extension PubspecUtils on Pubspec {
...
+  String root,
+  Iterable<String> patterns,
+) {
+  final members = <String>{};
+  for (final pattern in patterns) {
+    final normalized = p.posix.normalize(pattern.replaceAll(r'\', '/'));
+    if (looksLikeGlob(normalized)) {
+      try {
+        final glob = makeGlob(normalized);
+        for (final entity in glob.listSync(root: root)) {
```
engops-wk, line 232: This expansion will happily descend into hidden/generated trees, so a pattern like `packages/**` can pull packages out of `.dart_tool/`, `.symlinks/`, or `build/` and validate them. Please skip entities with a path segment starting with `.` (consistent with `listFilesWithExtensionIn` above) and pass `followLinks: false` to `listSync`.

lib/src/utils.dart
```diff
@@ -209,3 +210,39 @@ extension PubspecUtils on Pubspec {
 /// This function removes `./` paths and replaces all `\` with `/`.
 Glob makeGlob(String path) =>
     Glob(p.posix.normalize(path.replaceAll(r'\', '/')));
+
+/// Returns whether [path] looks like a glob pattern.
+bool looksLikeGlob(String path) => Glob.quote(path) != path;
```
engops-wk, line 215: Detecting globs via `Glob.quote(path) != path` depends on which characters that helper happens to escape, which isn't part of glob's API contract — a literal member dir like `my-package` or `pkg.v2` can be misclassified as a glob and then silently dropped if it doesn't match. Please use an explicit metacharacter check (e.g. `RegExp(r'[*?\[\]{}]').hasMatch(path)`) and add unit tests for `-`, `.`, and `,` literals.

lib/src/utils.dart
```diff
@@ -209,3 +210,39 @@ extension PubspecUtils on Pubspec {
...
+
+/// Resolves workspace member patterns to concrete package paths.
+///
+/// Glob patterns are expanded to directories beneath [root] that contain a
+/// `pubspec.yaml`, matching how `dart pub` resolves workspace members.
```
engops-wk, line 220: This doc comment asserts the behavior matches `dart pub`, but pub's `workspace:` entries are documented as plain relative paths — please verify and link the pub/Flutter source or issue (#176) that establishes glob support, or reword to state this is a dependency_validator-specific leniency and why "directory contains a pubspec.yaml" is the right filter.

test/utils.dart
```diff
@@ -99,7 +102,8 @@ Future<void> checkWorkspace({
         'dart_dependency_validator.yaml',
         jsonEncode(workspaceConfig.toJson()),
       ),
-    d.dir('subpackage', [
+    ...extraDirs,
+    d.dir(subpackagePath, [
```
engops-wk, line 106: The `extraDirs` descriptors are placed as siblings of the root `pubspec.yaml`, but tests pass `d.dir('packages/no_pubspec', [])` — a nested path in a single descriptor name. Confirm `test_descriptor` actually creates the nested directory here rather than a literal name; otherwise the "ignores directories without pubspec.yaml" test is vacuous and passes for the wrong reason.

lib/src/dependency_validator.dart
```diff
@@ -157,8 +163,8 @@ Future<bool> checkPackage({required String root}) async {
   final publicDirGlobs = [for (final dir in publicDirs) makeGlob('$dir**')];
 
   final subpackageGlobs = [
-    for (final subpackage in pubspec.workspace ?? [])
-      makeGlob('$root/$subpackage**'),
+    for (final subpackage in workspaceMembers)
+      makeGlob('${p.join(root, subpackage)}**'),
```
engops-wk, line 167: `p.join(root, subpackage)` returns a platform-separator path on Windows (`C:\ws\packages\foo`) and concatenating `**` yields `...\foo**`, which `makeGlob` then normalizes into a valid glob — but if the path has no trailing separator the glob also matches sibling dirs like `foo-extra`. Use `makeGlob(p.posix.join(...))` with an explicit `/` before `**` so the boundary is unambiguous, and add a test for a `packages/foo` + `packages/foo_extra` pair.

test/workspace_test.dart
```diff
@@ -149,4 +149,44 @@ void main() => group('Workspaces', () {
...
+          () => checkWorkspace(
+            workspace: [],
+            workspaceDeps: {},
+            workspaceMembers: ['packages/*'],
+            subpackagePath: 'packages/subpackage',
+            subpackage: usesHttp,
+            subpackageDeps: dependsOnHttp,
+          ),
+        );
+
```
engops-wk, line 165: These first two tests pass vacuously: if glob resolution returned an empty member list, `checkPackage` would validate only the clean root and still return `isTrue`. Add a positive assertion that the matched subpackage was actually visited (e.g. capture logs and assert on `Validating dependencies for subpackage`), so a regression in glob expansion actually fails these tests.

engops-wk (commented):
## Review — Changes Requested (round 1)

> Note: GitHub rejected a formal `REQUEST_CHANGES` event because the review is being submitted on my own PR, so this is posted as a comment. **Treat it as REQUEST_CHANGES — please do not merge until the items below are addressed.**

Thanks for tackling this — the shape of the change is right: resolve workspace members once, then use that resolved list for both recursion and the unused-file subpackage globs. The `p.join` cleanup over the old `'$root/$package'` string concatenation is a genuine improvement, and factoring resolution into a unit-testable function in `utils.dart` is the correct boundary.

I'm blocking on a few things that will bite us in production or hide regressions.

**Correctness / robustness**
1. `looksLikeGlob` is built on `Glob.quote(path) != path`. That's an inference over an implementation detail of `package:glob`'s quoting rules, not a documented contract. If `quote` escapes any character that is legal in a directory name (`-`, `.`, `,`, `{`), a literal member path gets routed down the glob branch and can be silently dropped. Highest-risk item, because the failure mode is "validation silently checks nothing."
2. Glob expansion has no hidden-directory or symlink guard. Every other listing helper in `utils.dart` deliberately filters segments starting with `.`; this one does not. A `packages/**` pattern will reach into `.dart_tool/`, Flutter's `.symlinks/`, and `build/` output.
3. Failures are silent. `on FormatException { /* skip */ }` and "glob matched zero packages" both produce an empty result with no diagnostic. Today a bad literal path produces a loud `pubspec.yaml not found` and returns `false`. Please preserve that signal — `checkPackage` already models it with `logger.shout(yellow.wrap('invalid glob syntax: "$s"'))`.
4. Subpackage glob construction concatenates `**` directly onto a joined path with no separator, so `packages/foo**` also matches a sibling `packages/foo_extra`, and on Windows the joined path uses `\` before normalization.

**Test strategy** — the part I'd most like strengthened. Two of the three new integration tests assert `isTrue` on a configuration whose root is clean, so they pass whether or not the glob resolved anything; only the `matcher: isFalse` case actually proves expansion happened. Please add a positive assertion that the matched subpackage was visited. Relatedly, confirm `d.dir('packages/no_pubspec', [])` actually creates a nested directory under `test_descriptor` rather than one literal-named entry — if it doesn't, that test asserts nothing.

**Docs** — the doc comment claims parity with `dart pub`. Pub documents `workspace:` entries as relative paths; if globs aren't officially supported there, say plainly that this is a deliberate dependency_validator leniency and link #176, so a future maintainer doesn't "fix" it back and break users.

None of this is architecturally wrong — it's about making failure modes loud and tests non-vacuous. Happy to re-review quickly once addressed.

PR #191, opened 2026-09-08
FEDX-7265: Dependencies used inside hook/ must be regular dependencies, not dev_dependencies

engops-wk:
Opened by @matthewnitschke-wk with the ai-sdlc workflow for FEDX-7265

## Motivation
Dependencies imported inside `hook/` directories run at install time by the package manager. If such a dependency is listed only under `dev_dependencies` (or not declared at all), it won't be available when a consuming package installs in production mode, causing hook execution to fail. `dependency_validator` did not previously check `hook/` usage against declared dependencies, allowing this misconfiguration to go undetected.

## Changes
- Added `hookDirs` handling in `lib/src/dependency_validator.dart`, scanning Dart/SCSS/Less files under `hook/` (in addition to the existing `publicDirs` scan for `bin/` and `lib/`).
- Excluded `hook/` from the "non-public" file scan so it isn't double-counted.
- Added a new check flagging packages used in `hook/` but not declared as any dependency ("missing hook dependencies").
- Extended the under-promoted dev_dependency check so packages used only in `hook/` and declared as `dev_dependency` are also flagged, requiring promotion to a regular dependency.
- Updated `README.md` to document the new "Under-promoted (hook)" validation case.
- Added a `CHANGELOG.md` entry describing the new validation.
- Added test coverage in `test/executable_test.dart` for: under-promoted hook dependencies (including ignore-list and deprecated pubspec-config paths), missing hook dependencies (including exclude and ignore paths), and a passing case with correctly declared hook dependencies.

## Testing/QA Instructions
- Run `dart test` and confirm all tests pass, including the new "fails when there are under promoted packages in hook/", "fails when there are packages missing from hook/", and "passes when hook dependencies are regular dependencies" groups.
- Manually verify: create a package with `hook/build.dart` importing a package declared only as a `dev_dependency`; run the validator and confirm it reports the package as needing promotion.
- Manually verify: create a package with `hook/build.dart` importing a package not declared at all; run the validator and confirm it reports the package as "used in hook/ but are not dependencies".
- Confirm that declaring the package as a regular dependency (or adding it to the `ignore` config) clears the failure and the validator reports "No dependency issues found!".
- CI passing on the added/updated tests is sufficient sign-off for this change.

## Jira ticket
FEDX-7265

## Checklist
- [ ] Changes follow the repository's coding standards
- [ ] Tests have been added/updated and pass locally
- [ ] Documentation (README/CHANGELOG) has been updated
- [ ] No breaking changes to public APIs
- [ ] Ready for code review

## Intent

This change extends the dependency_validator to treat the `hook/` directory as a public-facing directory for dependency resolution purposes. Dependencies imported in hook scripts run at install time in production mode, so they must be declared as regular dependencies rather than dev_dependencies or left undeclared. The validator now scans `hook/` for package imports and flags cases where packages are missing or only present as dev_dependencies.

## How To QA

1. 1. Run 'dart test test/executable_test.dart' and confirm all tests pass, including new groups: 'fails when there are under promoted packages in hook/', 'fails when there are packages missing from hook/', and 'passes when hook dependencies are regular dependencies'
2. 2. Create a test package with 'hook/build.dart' importing a package listed only as dev_dependency in pubspec.yaml; run 'dart run dependency_validator' and confirm it fails reporting the package needs promotion from dev_dependency to regular dependency
3. 3. Create a test package with 'hook/build.dart' importing a package not listed in pubspec.yaml at all; run validator and confirm it fails with message 'These packages are used in hook/ but are not dependencies:' listing that package
4. 4. Add the test package to regular dependencies (or add to ignore list in dependency_validator config); re-run validator and confirm it passes with 'No dependency issues found!'
5. 5. Verify README.md and CHANGELOG.md render correctly and describe the new hook-related validation check

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ used even if it isn't imported.
 
 - Missing: When a dependency is used in the package but not declared in the `pubspec.yaml`
 - Under-promoted: When a dependency is used within `lib/` but only declared as a dev_dependency.
+- Under-promoted (hook): When a dependency is used within `hook/` but only declared as a dev_dependency. Hook scripts run at install time and require regular dependencies.
 - Over-promoted: When a dependency is only used outside `lib/` but declared as a dependency.
 - Unused: When a dependency is not used in the package but declared in the `pubspec.yaml`.
 
--- a/lib/src/dependency_validator.dart
+++ b/lib/src/dependency_validator.dart
@@ -106,6 +106,7 @@ Future<bool> checkPackage({required String root}) async {
   );
 
   final publicDirs = ['$root/bin/', '$root/lib/'];
+  final hookDirs = ['$root/hook/'];
   logger.fine("Excluding: $excludes");
   final publicDartFiles = [
     for (final dir in publicDirs) ...listDartFilesIn(dir, excludes),
@@ -115,6 +116,15 @@ Future<bool> checkPackage({required String root}) async {
   ];
   final publicLessFiles = [
     for (final dir in publicDirs) ...listLessFilesIn(dir, excludes),
+  ];
+  final hookDartFiles = [
+    for (final dir in hookDirs) ...listDartFilesIn(dir, excludes),
+  ];
+  final hookScssFiles = [
+    for (final dir in hookDirs) ...listScssFilesIn(dir, excludes),
+  ];
+  final hookLessFiles = [
+    for (final dir in hookDirs) ...listLessFilesIn(dir, excludes),
   ];
 
   logger
@@ -129,6 +139,18 @@ Future<bool> checkPackage({required String root}) async {
     ..fine(
       'public facing less files:\n'
       '${bulletItems(publicLessFiles.map((f) => f.path))}\n',
+    )
+    ..fine(
+      'hook dart files:\n'
+      '${bulletItems(hookDartFiles.map((f) => f.path))}\n',
+    )
+    ..fine(
+      'hook scss files:\n'
+      '${bulletItems(hookScssFiles.map((f) => f.path))}\n',
+    )
+    ..fine(
+      'hook less files:\n'
+      '${bulletItems(hookLessFiles.map((f) => f.path))}\n',
     );
 
   // Read each file in lib/ and parse the package names from every import and
@@ -154,7 +176,29 @@ Future<bool> checkPackage({required String root}) async {
     '${bulletItems(packagesUsedInPublicFiles)}\n',
   );
 
+  final packagesUsedInHook = <String>{};
+  for (final file in hookDartFiles) {
+    packagesUsedInHook.addAll(getDartDirectivePackageNames(file));
+  }
+  for (final file in hookScssFiles) {
+    final matches = importScssPackageRegex.allMatches(file.readAsStringSync());
+    for (final match in matches) {
+      packagesUsedInHook.add(match.group(1)!);
+    }
+  }
+  for (final file in hookLessFiles) {
+    final matches = importLessPackageRegex.allMatches(file.readAsStringSync());
+    for (final match in matches) {
+      packagesUsedInHook.add(match.group(1)!);
+    }
+  }
+  logger.fine(
+    'packages used in hook/:\n'
+    '${bulletItems(packagesUsedInHook)}\n',
+  );
+
   final publicDirGlobs = [for (final dir in publicDirs) makeGlob('$dir**')];
+  final hookDirGlobs = [for (final dir in hookDirs) makeGlob('$dir**')];
 
   final subpackageGlobs = [
     for (final subpackage in pubspec.workspace ?? [])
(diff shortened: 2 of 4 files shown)


lib/src/dependency_validator.dart
```diff
@@ -166,16 +210,19 @@ Future<bool> checkPackage({required String root}) async {
   final nonPublicDartFiles = listDartFilesIn('$root/', [
     ...excludes,
     ...publicDirGlobs,
+    ...hookDirGlobs,
```
engops-wk, line 213: Excluding `hookDirGlobs` here removes hook/ imports from `packagesUsedOutsidePublicDirs`, but `unusedDependencies` still only subtracts `packagesUsedInPublicFiles` and `packagesUsedOutsidePublicDirs` — so any dependency used *only* in `hook/` is now falsely reported as unused. Add `.difference(packagesUsedInHook)` to the `unusedDependencies` computation.

test/executable_test.dart
```diff
@@ -150,6 +150,104 @@ void main() {
...
+      });
+
+      test('except when they are ignored', () async {
+        result = await checkProject(project: project, config: ignorePackages);
+        expect(result.exitCode, 0);
+      });
+    });
+
+    group('passes when hook dependencies are regular dependencies', () {
+      test('', () async {
```
engops-wk, line 237: This is the test the `unit-tests` CI job is failing on: `path` is only imported from `hook/build.dart`, so it lands in `unusedDependencies` and the run exits 1 instead of 0. After fixing the unused-dependency logic, also assert `result.stderr` does not contain 'may be unused' so this regression can't come back silently.

lib/src/dependency_validator.dart
```diff
@@ -242,11 +289,29 @@ Future<bool> checkPackage({required String root}) async {
     result = false;
   }
 
+  // Packages that are used in hook/ but are not dependencies.
+  final missingHookDependencies =
```
engops-wk, line 293: This block isn't `dart format` (3.7 tall-style) clean — with no leading comment the initializer collapses onto the `final missingHookDependencies = packagesUsedInHook` line, which is why the `checks` job is red. Run `dart format .` and commit the result.

README.md
```diff
@@ -23,6 +23,7 @@ used even if it isn't imported.
 
 - Missing: When a dependency is used in the package but not declared in the `pubspec.yaml`
 - Under-promoted: When a dependency is used within `lib/` but only declared as a dev_dependency.
+- Under-promoted (hook): When a dependency is used within `hook/` but only declared as a dev_dependency. Hook scripts run at install time and require regular dependencies.
```
engops-wk, line 26: Please also update the neighboring "Over-promoted" and "Unused" bullets to state that packages imported from `hook/` are exempt, and add a note to CHANGELOG.md that this validation can fail packages that previously passed (hook deps declared as dev_dependencies must be moved) so consumers aren't surprised on upgrade.

engops-wk (commented):

**Do not merge.** CI is red and there is a real correctness regression.

The direction here is right: `hook/` scripts run at install time, so their imports genuinely must be regular dependencies, and treating `hook/` as a distinct third bucket (rather than folding it into `publicDirs`) is the correct modeling choice — it keeps `bin/`+`lib/` semantics intact while giving hook usage its own promotion rule. The `overPromotedDependencies` exemption via `..removeAll(packagesUsedInHook)` shows you thought about rule interactions. One interaction was missed, and it's the one CI caught.

### Failing checks on `dfffdd5`

- **`unit-tests`** (Workiva/gha-dart-oss `test-unit.yaml@v0.1.14`, SDK 3.7.2) — the new group `passes when hook dependencies are regular dependencies` expects exit 0 / `No dependency issues found!`, but `path` is imported only from `hook/build.dart`. Because `hookDirGlobs` is now excluded from the non-public scan, `path` is in neither `packagesUsedInPublicFiles` nor `packagesUsedOutsidePublicDirs`, so it falls into `unusedDependencies` and the validator exits 1.
- **`checks`** (Workiva/gha-dart-oss `checks.yaml@v0.1.14`) — `dart format` verification fails; the new `missingHookDependencies` cascade block is not in the shape the 3.7 formatter emits.

### Blocking issues

1. **Hook-only dependencies are falsely reported as unused** (`lib/src/dependency_validator.dart`, `unusedDependencies`). Root cause of the test failure and a user-facing bug: a package that correctly declares its hook dependency as a regular dependency will be told to delete it. Needs `.difference(packagesUsedInHook)`.
2. **Formatting** — run `dart format .` and commit.
3. **Docs incomplete** — README/CHANGELOG describe the new category but not the resulting exemptions (over-promoted, unused), and CHANGELOG doesn't flag the potentially-breaking upgrade impact.

### Non-blocking observations

- Add a one-line comment above `hookDirs` explaining *why* `hook/` is special (install-time execution). Future maintainers won't have the FEDX-7265 context.
- Scanning `hook/` for `.scss`/`.less` is harmless and symmetric but practically dead; fine to keep, just noting the tradeoff.
- Once fixed, add a test asserting a hook-only *regular* dependency is not flagged as over-promoted, to lock in the other half of the new behavior.

Please push fixes and re-request review once `build`, `checks`, and `unit-tests` are all green.

PR #192, opened 2026-09-08
FEDX-7269: Fix parse failures on final parameter modifiers with analyzer 13

engops-wk:
Opened by @matthewnitschke-wk with the ai-sdlc workflow for FEDX-7269

## Motivation
`getDartDirectivePackageNames` calls `parseString` without specifying a feature set, so it defaults to `FeatureSet.latestLanguageVersion()`. With analyzer 13 (picked up in dependency_validator 5.0.6), that "latest" version is unreleased and rejects `final` on formal parameters, even though the SDK analyzer accepts it fine. This breaks any project using `freezed`, since it emits `final` in generated constructors — several packages in downstream monorepos failed CI overnight with no source change, because CI installs this tool unpinned.

## Changes
- Added `featureSetForSdkConstraint` in `lib/src/import_export_ast_visitor.dart`, which derives a `FeatureSet` from a package's SDK constraint's minimum version, the same way the analyzer determines language version internally. Falls back to `FeatureSet.latestLanguageVersion()` when the constraint is missing or unbounded.
- `getDartDirectivePackageNames` now accepts an optional `featureSet` parameter and forwards it to `parseString`.
- `checkPackage` in `lib/src/dependency_validator.dart` computes the feature set from `pubspec.environment['sdk']` once and passes it to both calls to `getDartDirectivePackageNames` (public and non-public Dart files).
- Added a regression test in `test/executable_test.dart` covering a file with `final` parameter modifiers.
- Updated `CHANGELOG.md`.

## Testing/QA Instructions
- Run `dart test test/executable_test.dart` and confirm all tests pass, including the new test verifying files with `final` parameter modifiers parse without error and report "No dependency issues found!".
- Manually confirm packages with a normal SDK constraint continue to validate dependencies correctly, and packages without an SDK constraint fall back to the previous latest-language-version behavior.

## Jira ticket
FEDX-7269

## Intent

Fixes a parsing issue where `getDartDirectivePackageNames` defaults to an unreleased language version that rejects `final` on formal parameters, breaking packages using `freezed`. This change derives the feature set from the package's SDK constraint instead, so parsing respects the package's declared language version rather than assuming the newest possible version.

## How To QA

1. Run dart test test/executable_test.dart and confirm all tests pass, including the new regression test for final parameter modifiers
2. Manually create a test project with final formal parameters (e.g., void log(final Logger logger, {final String message = ''}) {...}) and run the dependency validator, confirming exit code 0 and 'No dependency issues found!' output
3. Verify packages with normal SDK constraints in pubspec.yaml still correctly detect used/unused dependencies
4. Confirm fallback behavior: packages without SDK constraints or unbounded constraints still fall back to FeatureSet.latestLanguageVersion() without regressions

### DIFF
--- a/lib/src/import_export_ast_visitor.dart
+++ b/lib/src/import_export_ast_visitor.dart
@@ -1,16 +1,37 @@
 import 'dart:io';
 
+import 'package:analyzer/dart/analysis/features.dart';
 import 'package:analyzer/dart/analysis/results.dart';
 import 'package:analyzer/dart/analysis/utilities.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:pub_semver/pub_semver.dart';
+
+/// Builds the feature set matching the language version a package declares,
+/// the same way the analyzer derives it from the SDK constraint.
+///
+/// Without this, [parseString] falls back to the newest language version the
+/// analyzer knows about, which may be unreleased and reject valid code.
+FeatureSet featureSetForSdkConstraint(VersionConstraint? sdkConstraint) {
+  final min = sdkConstraint is VersionRange ? sdkConstraint.min : null;
+  if (min == null) return FeatureSet.latestLanguageVersion();
+
+  return FeatureSet.fromEnableFlags2(
+    sdkLanguageVersion: Version(min.major, min.minor, 0),
+    flags: const [],
+  );
+}
 
 /// Returns the list of package names that are exported and imported into the
 /// provided dart file
-Set<String> getDartDirectivePackageNames(File file) {
+Set<String> getDartDirectivePackageNames(File file, {FeatureSet? featureSet}) {
   ParseStringResult parsed;
   try {
-    parsed = parseString(content: file.readAsStringSync(), path: file.path);
+    parsed = parseString(
+      content: file.readAsStringSync(),
+      path: file.path,
+      featureSet: featureSet,
+    );
   } on ArgumentError catch (e) {
     print('Error parsing: ${file.path}');
     print(e.message);
--- a/lib/src/import_export_ast_visitor.dart
+++ b/lib/src/import_export_ast_visitor.dart
@@ -7,40 +7,81 @@ import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
 import 'package:pub_semver/pub_semver.dart';
 
+/// Returns the SDK language version implied by [sdkConstraint], or null when
+/// the latest language version should be used.
+Version? languageVersionForSdkConstraint(VersionConstraint? sdkConstraint) {
+  final min = sdkConstraint is VersionRange ? sdkConstraint.min : null;
+  if (min == null) return null;
+
+  return Version(min.major, min.minor, 0);
+}
+
 /// Builds the feature set matching the language version a package declares,
 /// the same way the analyzer derives it from the SDK constraint.
 ///
 /// Without this, [parseString] falls back to the newest language version the
 /// analyzer knows about, which may be unreleased and reject valid code.
 FeatureSet featureSetForSdkConstraint(VersionConstraint? sdkConstraint) {
-  final min = sdkConstraint is VersionRange ? sdkConstraint.min : null;
-  if (min == null) return FeatureSet.latestLanguageVersion();
+  final languageVersion = languageVersionForSdkConstraint(sdkConstraint);
+  if (languageVersion == null) return FeatureSet.latestLanguageVersion();
 
   return FeatureSet.fromEnableFlags2(
-    sdkLanguageVersion: Version(min.major, min.minor, 0),
+    sdkLanguageVersion: languageVersion,
     flags: const [],
   );
 }
 
 /// Returns the list of package names that are exported and imported into the
 /// provided dart file
-Set<String> getDartDirectivePackageNames(File file, {FeatureSet? featureSet}) {
+Set<String> getDartDirectivePackageNames(
+  File file, {
+  FeatureSet? featureSet,
+  Version? derivedLanguageVersion,
+}) {
   ParseStringResult parsed;
+  final content = file.readAsStringSync();
   try {
     parsed = parseString(
-      content: file.readAsStringSync(),
+      content: content,
       path: file.path,
       featureSet: featureSet,
     );
   } on ArgumentError catch (e) {
-    print('Error parsing: ${file.path}');
-    print(e.message);
-    exit(1);
+    if (derivedLanguageVersion != null) {
+      try {
+        parsed = parseString(
+          content: content,
+          path: file.path,
+          featureSet: FeatureSet.latestLanguageVersion(),
+        );
+      } on ArgumentError catch (retryError) {
+        _reportParseError(
+          file,
+          derivedLanguageVersion: derivedLanguageVersion,
+          error: retryError,
+        );
+      }
+    } else {
+      _reportParseError(file, error: e);
+    }
   }
 
   final visitor = ImportExportVisitor();
   parsed.unit.visitChildren(visitor);
   return visitor.packageNames;
+}
+
+Never _reportParseError(
+  File file, {
+  Version? derivedLanguageVersion,
+  required ArgumentError error,
+}) {
+  print('Error parsing: ${file.path}');
+  if (derivedLanguageVersion != null) {
+    print('Derived language version: $derivedLanguageVersion');
+  }
+  print(error.message);
+  exit(1);
 }
 
 class ImportExportVisitor extends GeneralizingAstVisitor {
--- a/test/import_export_ast_visitor_test.dart
+++ b/test/import_export_ast_visitor_test.dart
@@ -1,0 +1,113 @@
+// Copyright 2017 Workiva Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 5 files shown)


lib/src/import_export_ast_visitor.dart
```diff
@@ -1,16 +1,37 @@
...
 /// provided dart file
-Set<String> getDartDirectivePackageNames(File file) {
+Set<String> getDartDirectivePackageNames(File file, {FeatureSet? featureSet}) {
   ParseStringResult parsed;
   try {
-    parsed = parseString(content: file.readAsStringSync(), path: file.path);
+    parsed = parseString(
+      content: file.readAsStringSync(),
+      path: file.path,
+      featureSet: featureSet,
```
engops-wk, line 47: Passing a lower, SDK-constraint-derived feature set means files whose syntax exceeds the declared min SDK now produce diagnostics, so `parseString` throws and we `exit(1)` — a new CI break for repos that pass today. Please retry once with `FeatureSet.latestLanguageVersion()` on `ArgumentError` before failing, and include the derived language version in the error output.

lib/src/import_export_ast_visitor.dart
```diff
@@ -1,16 +1,37 @@
...
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:pub_semver/pub_semver.dart';
+
+/// Builds the feature set matching the language version a package declares,
+/// the same way the analyzer derives it from the SDK constraint.
+///
+/// Without this, [parseString] falls back to the newest language version the
+/// analyzer knows about, which may be unreleased and reject valid code.
+FeatureSet featureSetForSdkConstraint(VersionConstraint? sdkConstraint) {
```
engops-wk, line 24: This function now decides parse behavior for every file in every consuming repo but has no direct test coverage — the null/unbounded/exact-version branches are only manually verified. Please add unit tests covering `null`, `VersionConstraint.any`, an exact version (`3.6.0`), and a normal range.

engops-wk (commented):

The diagnosis is correct and the fix is directionally the right one: parsing at the package's declared language version instead of `FeatureSet.latestLanguageVersion()` is exactly how the analyzer/SDK behave, and it unblocks the `freezed`/`final`-parameter breakage. `checkPackage` computing the feature set once per package (including per sub-package in the workspace recursion) is the right placement, and I verified `test/utils.dart` writes `environment: sdk: ^3.6.0` into every generated project, so the new regression test does genuinely exercise the derived feature set.

Two things need to change before merge.

**1. This can now hard-fail repos that previously passed, with an unhelpful message.**
Deriving from the SDK constraint *lowers* the language version, so any package whose declared min SDK doesn't cover the syntax it actually contains (a misdeclared `environment:` bound, vendored/generated files, third-party code under `tool/`) will now produce diagnostics, hit the `on ArgumentError` branch in `getDartDirectivePackageNames`, print `Error parsing: <path>` and `exit(1)`. That's a brand-new CI-breaking mode for a tool the ticket explicitly says is installed **unpinned** across downstream monorepos — we'd trade one overnight breakage for a different one. Please add a retry with `FeatureSet.latestLanguageVersion()` before failing (strictly no worse than today's behavior in either direction) and include the derived language version in the error output so the cause is diagnosable without reading this source file.

**2. `featureSetForSdkConstraint` has branchy logic and zero direct tests.**
It now gates parsing for every file in every consuming repo, and the PR description marks the fallback paths as "manually confirmed." That's the wrong level of assurance for this blast radius. Please add unit tests for the null / unbounded / exact-version / normal-range cases.

### Non-blocking notes (your call)
- **Placement:** `featureSetForSdkConstraint` is an SDK-constraint utility, not an AST-visitor concern. `lib/src/utils.dart` already imports `pub_semver` and `pubspec_parse` and hosts exactly this kind of helper, with `test/utils_test.dart` as the natural home for its tests. Moving it would improve cohesion and discoverability.
- **`includeMin: false`:** `sdk: '>3.6.0'` yields language version 3.6 even though 3.6.0 itself is excluded. Vanishingly rare and matches what the analyzer effectively does; I'd leave it.
- **Release mechanics:** the motivation depends on publishing, and `pubspec.yaml` still says `version: 5.0.6` with the entry under `# Unreleased`. Please confirm the auto-release tooling bumps the version, otherwise consumers on unpinned installs never pick this up.
- The CHANGELOG entry is clear and accurately scoped — written from the consumer's perspective rather than describing the code change. Nice.

test/import_export_ast_visitor_test.dart
```diff
@@ -0,0 +1,113 @@
...
+  });
+
+  group('getDartDirectivePackageNames', () {
+    test('retries with latest language version on parse failure', () {
+      final file = File('${Directory.systemTemp.path}/retry_test.dart')
+        ..writeAsStringSync('''
+import 'package:logging/logging.dart';
+
+void log(List<int>? values, final Logger logger) {
+  final copied = [?values];
```
engops-wk, line 95: Null-aware elements (`[?values]`) require language 3.8, but CI pins SDK 3.7.2, so the analyzer resolved there tops out at 3.7 and the retry parse will also throw — which lands in `_reportParseError` and calls `exit(1)`, aborting the whole test process rather than failing one test. Please use syntax that is newer than the derived version but supported by the pinned SDK (e.g. derive from `^2.12.0` and use a Dart 3 record/pattern), or exercise the retry through `checkProject` in a subprocess.

test/import_export_ast_visitor_test.dart
```diff
@@ -0,0 +1,113 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+@TestOn('vm')
+import 'dart:io';
+
+import 'package:analyzer/dart/analysis/features.dart';
+import 'package:analyzer/src/dart/analysis/experiments.dart';
```
engops-wk, line 19: Importing `package:analyzer/src/...` violates the `implementation_imports` lint that ships in `workiva_analysis_options/v2.yaml`, so the `checks` job will fail; it also couples the test to analyzer internals (`ExperimentStatus`, `getSdkLanguageVersion_forTesting`) across a `>=7.1.0 <15.0.0` range — exactly the drift that caused this bug. Please assert behavior through the public API instead (e.g. `featureSet.isEnabled(Feature.<x>)` or parse a snippet that only a newer version accepts).

lib/src/import_export_ast_visitor.dart
```diff
@@ -1,27 +1,89 @@
...
+      path: file.path,
+      featureSet: featureSet,
+    );
   } on ArgumentError catch (e) {
-    print('Error parsing: ${file.path}');
-    print(e.message);
-    exit(1);
+    if (derivedLanguageVersion != null) {
+      try {
+        parsed = parseString(
```
engops-wk, line 52: The silent fallback to `latestLanguageVersion()` means a real parse failure at the declared version is invisible, and the only signal a downgraded feature set was in play is the version line printed on total failure. Please log (at `logger.fine`/`warning`) when the retry succeeds so the behavior is diagnosable in the field.

lib/src/import_export_ast_visitor.dart
```diff
@@ -1,27 +1,89 @@
...
+    flags: const [],
+  );
+}
 
 /// Returns the list of package names that are exported and imported into the
 /// provided dart file
-Set<String> getDartDirectivePackageNames(File file) {
+Set<String> getDartDirectivePackageNames(
+  File file, {
+  FeatureSet? featureSet,
```
engops-wk, line 38: `derivedLanguageVersion` is redundant with `featureSet` — it only acts as a "retry is allowed" flag, so a caller can pass a derived `featureSet` with a null version (or vice versa) and get inconsistent behavior. Please collapse this into a single parameter (e.g. take `VersionConstraint? sdkConstraint` and derive both internally) so the API can't be misused.

engops-wk (commented):
## Review round 2 — changes requested (posted as COMMENT; GitHub blocks REQUEST_CHANGES on one's own PR)

**Verdict: not ready to merge.** 4 blocking items, 2 of which will fail CI.

The core fix is the right call: deriving the `FeatureSet` from the package's declared SDK constraint instead of letting `parseString` default to an unreleased "latest" language version is the correct model, it matches how the analyzer itself derives language version, and the retry-with-latest safety net sensibly guards the inverse failure mode (a package whose code is newer than its floor constraint). `checkPackage` computing this once per package and the `executable_test.dart` regression test are both good.

### Blocking

1. **`test/import_export_ast_visitor_test.dart` imports `package:analyzer/src/dart/analysis/experiments.dart`** — an implementation import. `analysis_options.yaml` includes `workiva_analysis_options/v2.yaml`, which enables `implementation_imports`, so the `checks` job fails. It also casts `FeatureSet` to the internal `ExperimentStatus` and calls `getSdkLanguageVersion_forTesting`, pinning the suite to analyzer internals across a `>=7.1.0 <15.0.0` dependency range — precisely the drift that caused this bug. Assert via the public API instead.

2. **The retry test uses null-aware elements (`[?values]`), which require language 3.8, but `.github/workflows/ci.yaml` pins SDK 3.7.2.** The retry parse at `latestLanguageVersion()` throws there too, which routes into `_reportParseError` → `exit(1)`, killing the test *process* rather than failing one test. Pick syntax newer than the derived version but supported by the pinned SDK, or drive the retry through `checkProject` in a subprocess.

3. **`getDartDirectivePackageNames(file, {featureSet, derivedLanguageVersion})` takes two parameters that must agree**; `derivedLanguageVersion` is effectively just a "retry allowed" flag, and a caller can pass one without the other. Collapse to a single input.

4. **The retry succeeds silently.** This tool runs unpinned in CI across many repos; emit a log line when the fallback engages so the next parse discrepancy is diagnosable.

Nothing here is a rethink of the approach — fix the test's analyzer coupling and SDK assumption, tighten the signature, add the log line, and this is good to go.

PR #193, opened 2026-09-08
FEDX-7267: Support Dart doc imports in dependency_validator

engops-wk:
Opened by @matthewnitschke-wk with the ai-sdlc workflow for FEDX-7267

# Pull Request

## Motivation
`dependency_validator` didn't recognize Dart 3.8's `@docImport` tags in documentation comments as a form of package usage. Packages referenced only through a doc import in `lib/` were flagged as unused, even though they were legitimately referenced (for doc-linking purposes only, not runtime). Since doc imports don't create a runtime dependency, packages referenced this way in `lib/` should be treated as candidates for `dev_dependencies` rather than regular `dependencies`.

## Changes
- **`lib/src/import_export_ast_visitor.dart`**: Introduced `DartPackageUsage`, a container holding both `directivePackageNames` (from real `import`/`export` directives) and `docImportPackageNames` (from `@docImport` tags in doc comments). Added `getDartPackageUsage(File)` which parses a file and returns both sets. `ImportExportVisitor` now overrides `visitComment` to extract `@docImport` URIs, and `visitDirective` was tightened to only treat top-level `CompilationUnit` directives as real imports/exports (avoiding false positives from nested directive-like nodes). `getDartDirectivePackageNames` is retained as a thin wrapper over the new API for compatibility.

- **`lib/src/dependency_validator.dart`**: Wired the new doc-import tracking into the dependency analysis. For both public (`lib/`) and non-public files, the code now collects `docImportPackageNames` alongside directive-based usage, logs them separately at `fine` level, and folds doc-import-only packages from `lib/` into the "used outside lib/" set so they're evaluated as `dev_dependencies` candidates rather than runtime `dependencies`. This affects the missing-dependency, missing-dev-dependency, and over-promoted-dependency checks.

- **`CHANGELOG.md`**: Added an entry describing the new `@docImport` support and the dev-dependency treatment.

- **`test/executable_test.dart`**: Added a `doc imports` test group covering: passing when a package is used only via doc import in `lib/` and declared as a dev dependency; failing when such a package is missing from the pubspec entirely; not flagging under-promotion when the package is correctly a dev dependency; and failing (over-promoted) when the package is declared as a regular dependency instead of a dev dependency.

- **`test/import_export_ast_visitor_test.dart`** (new): Unit tests for `getDartPackageUsage` covering top-level import/export collection, ignoring nested/non-directive imports, collecting doc imports from library-level and declaration-level comments, collecting multiple doc imports, and ignoring `dart:` and relative doc imports.

## Testing/QA
- [ ] Full test suite passes (`dart test`)
- [ ] New test group `doc imports` in `test/executable_test.dart` passes
- [ ] New test file `test/import_export_ast_visitor_test.dart` passes
- [ ] Manual verification: package with doc import in `lib/` and dev_dependency declaration shows no issues
- [ ] Manual verification: same package in dependencies now reports as over-promoted
- [ ] Manual verification: same package missing from pubspec reports as missing from dev_dependencies
- [ ] Verbose logging shows fine-level doc import package usage lines

## Checklist
- [x] Code follows project style guidelines
- [x] New tests added for doc import functionality
- [x] CHANGELOG.md updated
- [x] Backward compatibility maintained (getDartDirectivePackageNames wrapper retained)
- [x] Documentation comments are clear and complete

## Intent

This change adds support for Dart 3.8's `@docImport` tags in documentation comments to the dependency_validator tool. Previously, packages referenced only through doc imports were incorrectly flagged as unused dependencies. The fix recognizes doc imports as legitimate package usage and treats doc-import-only packages from `lib/` as candidates for `dev_dependencies` rather than runtime `dependencies`, implementing proper categorization across the missing-dependency, missing-dev-dependency, and over-promoted-dependency checks.

## How To QA

1. 1. Execute full test suite using `dart test` command and verify all tests pass, including the new doc imports test group in test/executable_test.dart and the new test/import_export_ast_visitor_test.dart file
2. 2. Create a scratch package with a lib/ file containing `/// @docImport 'package:yaml/yaml.dart';` followed by `library;`, declare yaml as dev_dependency only, run dependency_validator and verify no issues are reported
3. 3. Move yaml to dependencies instead and re-run dependency_validator; verify it now reports yaml should be downgraded to dev_dependencies
4. 4. Remove yaml entirely from pubspec and re-run; verify dependency_validator reports yaml as missing from dev_dependencies
5. 5. Run with verbose logging enabled to confirm fine-level log lines for doc import package usage appear for both public and non-public file scans


lib/src/import_export_ast_visitor.dart
```diff
@@ -18,26 +37,40 @@ Set<String> getDartDirectivePackageNames(File file) {
...
-    if (node is! UriBasedDirective) return;
+    if (node.parent is CompilationUnit && node is UriBasedDirective) {
+      _addPackageName(node.uri.stringValue, packageNames);
+    }
+    node.visitChildren(this);
+  }
+
+  @override
+  void visitComment(Comment node) {
+    for (final docImport in node.docImports) {
```
engops-wk, line 61: `Comment.docImports` / `DocImport` are newer analyzer APIs, but the pubspec still allows `analyzer: ">=7.1.0"`. Please verify the minimum analyzer version that exposes `docImports` and raise the lower bound, otherwise consumers resolving an older analyzer get a compile error.

lib/src/import_export_ast_visitor.dart
```diff
@@ -5,9 +5,28 @@ import 'package:analyzer/dart/analysis/utilities.dart';
...
+/// provided dart file.
+Set<String> getDartDirectivePackageNames(File file) =>
+    getDartPackageUsage(file).directivePackageNames;
+
+/// Returns package names referenced via import/export directives and doc
+/// imports in the provided dart file.
+DartPackageUsage getDartPackageUsage(File file) {
   ParseStringResult parsed;
   try {
     parsed = parseString(content: file.readAsStringSync(), path: file.path);
```
engops-wk, line 32: `parseString` defaults to `FeatureSet.latestLanguageVersion()` of the *running* SDK, and `@docImport` requires the Dart 3.8 language feature — CI pins 3.7.2 (`.tool-versions`, `ci.yaml`), so doc imports parse as plain comment text and the new tests can't pass. Bump `.tool-versions`/`ci.yaml` to >= 3.8 and raise the pubspec `sdk` floor (or pass an explicit `featureSet`).

test/executable_test.dart
```diff
@@ -201,6 +201,116 @@ void main() {
       );
     });
 
+    group('doc imports', () {
+      test('passes when package is only referenced via doc import in lib/', () async {
```
engops-wk: This line is ~86 chars and the `isNot(contains(...))` blocks below aren't in tall-style form, so `dart format --set-exit-if-changed` (the `checks` job) fails. Run `dart format .` and commit the result.

lib/src/dependency_validator.dart
```diff
@@ -220,6 +230,14 @@ Future<bool> checkPackage({required String root}) async {
     'packages used outside public dirs:\n'
     '${bulletItems(packagesUsedOutsidePublicDirs)}\n',
   );
+  logger.fine(
+    'packages used via doc imports outside public dirs:\n'
+    '${bulletItems(docImportPackagesUsedOutsidePublicDirs)}\n',
+  );
+
+  final packagesUsedViaDocImport = docImportPackagesUsedInPublicFiles.union(
```
engops-wk, line 240: `docImportPackagesUsedOutsidePublicDirs` only feeds the unused-dependency check; it is never unioned into `packagesUsedOutsidePublicDirs`, so a package doc-imported *only* in `test/`/`tool/` is silently not reported as a missing dev_dependency (and a regular `dependencies` entry used only that way isn't flagged over-promoted). Fold it into the "used outside lib/" set so doc imports are treated consistently regardless of directory.

test/import_export_ast_visitor_test.dart
```diff
@@ -0,0 +1,74 @@
+import 'dart:io';
```
engops-wk, line 16: This unit test file bypasses `test/utils.dart` and depends on `d.sandbox` path construction directly; also it has no license header (all other files in the repo carry the Apache header) and no `@TestOn('vm')` annotation even though it uses `dart:io`. Add the header and `@TestOn('vm')` to keep the file consistent with the rest of `test/`.

test/import_export_ast_visitor_test.dart
```diff
@@ -0,0 +1,74 @@
...
+''');
+      expect(usage.directivePackageNames, {'foo', 'bar'});
+      expect(usage.docImportPackageNames, isEmpty);
+    });
+
+    test('ignores nested import directives', () async {
+      final usage = await parse('''
+import 'package:foo/foo.dart';
+
+void main() {
```
engops-wk: The test name says "ignores nested import directives" but the body only contains a comment inside `main()` — it doesn't exercise the new `node.parent is CompilationUnit` guard at all. Replace with a real nested case (e.g. an `import` inside an augmentation/part-like construct or a string that produces a non-top-level `UriBasedDirective`) or drop the test so it doesn't give false coverage confidence.

lib/src/import_export_ast_visitor.dart
```diff
@@ -18,26 +37,40 @@ Set<String> getDartDirectivePackageNames(File file) {
...
+    docImportPackageNames: visitor.docImportPackageNames,
+  );
 }
 
-class ImportExportVisitor extends GeneralizingAstVisitor {
+class ImportExportVisitor extends GeneralizingAstVisitor<void> {
   Set<String> packageNames = {};
+  Set<String> docImportPackageNames = {};
 
   @override
```
engops-wk, line 51: `visitDirective` now calls `node.visitChildren(this)` but `visitComment` does not, and `GeneralizingAstVisitor` overrides that skip `super`/`visitChildren` stop traversal. Confirm doc comments nested under directives (e.g. a `///` block on an `import`) are still reached, and add `node.visitChildren(this)` to `visitComment` for symmetry.

lib/src/dependency_validator.dart
```diff
@@ -244,8 +262,10 @@ Future<bool> checkPackage({required String root}) async {
 
   // Packages that are used outside lib/ but are not dev_dependencies.
   final missingDevDependencies =
-      // Start with packages _only_ used outside lib/
+      // Start with packages _only_ used outside lib/, and packages only
+      // referenced via `@docImport` in lib/.
       packagesUsedOutsidePublicDirs
```
engops-wk, line 267: Reporting a `lib/`-only doc import under the message "These packages are used **outside lib/** but are not dev_dependencies" (and "only used outside lib/ ... downgraded") is misleading — the package *is* referenced in `lib/`, just in a doc comment. Please add doc-import-specific messaging (or amend these strings) so users can act on the output without guessing.

engops-wk (commented):

Thanks for tackling this — the core idea (treat `@docImport`-only usage in `lib/` as a dev-dependency signal rather than a runtime dependency) is the right call, and the `DartPackageUsage` value object keeps the API honest while preserving `getDartDirectivePackageNames` for compatibility. The direction is sound; the execution has blocking gaps. **Do not merge.**

### CI is red — this cannot merge as-is
Failing checks on `083bfc1`:
- **`checks` (dart analyze + `dart format --set-exit-if-changed`)** — `test/executable_test.dart` has unformatted additions: the `passes when package is only referenced via doc import in lib/` test declaration exceeds the line limit, and the `isNot(contains(...))` argument blocks are not in tall-style form.
- **`unit-tests` (`dart test`)** — the new `doc imports` group in `test/executable_test.dart` and all of `test/import_export_ast_visitor_test.dart` cannot pass on the pinned SDK. `.tool-versions` and `.github/workflows/ci.yaml` both pin **Dart 3.7.2**, but `@docImport` is a **Dart 3.8** language feature. `parseString` defaults to the running SDK's latest language version, so on 3.7.2 `@docImport` is parsed as ordinary comment text and `Comment.docImports` is empty.
- **`build`** — fails for the same version-constraint reason: `pubspec.yaml` still allows `analyzer: ">=7.1.0 <15.0.0"`, and the `Comment.docImports` / `DocImport` API postdates that floor.

Please push fixes, re-run CI, and confirm green before the next round. I'm not approving on inspection alone given the SDK feature-gating risk.

### Blocking correctness/design issues
1. **SDK/analyzer version gating** — the feature is silently a no-op below Dart 3.8, and the analyzer lower bound is too low for `Comment.docImports`.
2. **`docImportPackagesUsedOutsidePublicDirs` is collected and logged but never folded into `packagesUsedOutsidePublicDirs`** — a package doc-imported only in `test/`/`tool/` is not reported as a missing dev_dependency, and a regular `dependencies` entry used only that way isn't flagged over-promoted. Asymmetric with the `lib/` handling and untested.
3. **Misleading output messaging** — a `lib/`-only doc import is reported as "used outside lib/", which is factually wrong from the user's perspective and will generate support questions.
4. **Test quality** — `ignores nested import directives` does not exercise the new `node.parent is CompilationUnit` guard at all; the new test file is missing the Apache license header and `@TestOn('vm')` that the rest of `test/` carries.

### Please also confirm
- Does `visitComment` need `node.visitChildren(this)` for consistent traversal?
- README documentation for the new behavior — users need to know why a doc-imported package is now expected in `dev_dependencies`.
- The PR body's Testing/QA checklist is entirely unchecked. Complete it or state explicitly what was verified.

Happy to re-review quickly once CI is green.

PR #196, opened 2026-09-08
FEDX-7269: Fix parse failures on final parameter modifiers with analyzer 13

engops-wk:
Opened by Dustin Pauze with the ai-sdlc workflow for FEDX-7269

## Intent

Fix language feature set resolution in `dependency_validator` by deriving the feature set from the package's SDK constraint rather than defaulting to the latest language version. This prevents validation failures when projects use `final` parameters (as generated by `freezed`) and CI pulls in a newer analyzer version where "latest" language features are unsupported, allowing packages to be parsed against their actual target language version.

## Changes

- **lib/src/import_export_ast_visitor.dart**: Added `featureSetForSdkConstraint` function that maps `VersionConstraint` (pinned version, range, or unbounded/missing) to `FeatureSet` built from the constraint's minimum major/minor version, falling back to `FeatureSet.latestLanguageVersion()` when the constraint is null or has no resolvable minimum. Updated `getDartDirectivePackageNames` to accept an optional `featureSet` parameter, pass it to `parseString`, and retry with the latest feature set if the pinned one fails to parse before erroring out.

- **lib/src/dependency_validator.dart**: Updated `checkPackage` to accept a `workspaceSdkConstraint` parameter, resolve the effective SDK constraint for a package (its own `environment['sdk']`, or the inherited workspace root constraint when the package is part of a workspace and declares none), compute the corresponding `featureSet`, and pass it down through recursive workspace calls and both `getDartDirectivePackageNames` call sites (public and non-public Dart files).

- **CHANGELOG.md**: Added an entry describing the fix.

- **test/import_export_ast_visitor_test.dart** (new): Adds unit tests for `featureSetForSdkConstraint` covering bounded (`^3.6.0`), pinned (`3.6.0`), unbounded (`any`), null, and pre-release (`3.7.0-dev`) constraints, plus a test verifying `getDartDirectivePackageNames` falls back to the latest feature set when the pinned one fails to parse.

- **test/utils.dart**: Added `omitSubpackageEnvironment` flag to `checkWorkspace` test helper to build a subpackage pubspec without an `environment` key, for testing SDK constraint inheritance.

- **test/workspace_test.dart**: Adds a test verifying a sub-package without its own SDK environment inherits the root workspace's SDK constraint and correctly parses `final` parameters.

## How To QA

1. Run the full test suite with `dart test` and confirm all tests pass, including the new `import_export_ast_visitor_test.dart` and the added workspace test in `workspace_test.dart`.

2. Run `dart test test/import_export_ast_visitor_test.dart` and verify `featureSetForSdkConstraint` correctly resolves bounded/pinned constraints to expected major.minor versions, `any`/null constraints fall back to latest, and pre-release constraints map to their major.minor version.

3. Run `dart test test/workspace_test.dart` and confirm the new inheritance test passes for a sub-package with no `environment` key containing `final` formal parameter.

4. Create a manual test package with an SDK constraint (e.g. `sdk: '>=3.6.0 <4.0.0'`) and a `final` formal parameter in a public/exported location, run `dependency_validator` against it, and confirm no parse failure occurs.

5. Verify `CHANGELOG.md` contains an entry for the fix under the `# Unreleased` section.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Testing

- [x] New tests added
- [x] Existing tests updated
- [x] All tests passing

## Checklist

- [x] Code follows the style guidelines of this project
- [x] Self-review of own code completed
- [x] Comments added for complex logic
- [x] Documentation updated
- [x] No new warnings generated
- [x] Tests added/updated and all passing

### DIFF
--- a/lib/src/dependency_validator.dart
+++ b/lib/src/dependency_validator.dart
@@ -20,6 +20,7 @@ import 'package:io/ansi.dart';
 import 'package:logging/logging.dart';
 import 'package:package_config/package_config.dart';
 import 'package:path/path.dart' as p;
+import 'package:pub_semver/pub_semver.dart';
 import 'package:pubspec_parse/pubspec_parse.dart';
 
 import 'constants.dart';
@@ -27,7 +28,10 @@ import 'pubspec_config.dart';
 import 'utils.dart';
 
 /// Check for missing, under-promoted, over-promoted, and unused dependencies.
-Future<bool> checkPackage({required String root}) async {
+Future<bool> checkPackage({
+  required String root,
+  VersionConstraint? workspaceSdkConstraint,
+}) async {
   var result = true;
   if (!File('$root/pubspec.yaml').existsSync()) {
     logger.shout(red.wrap('pubspec.yaml not found'));
@@ -79,18 +83,27 @@ Future<bool> checkPackage({required String root}) async {
     sourceUrl: pubspecFile.uri,
   );
 
+  final rootSdkConstraint = pubspec.isWorkspaceRoot
+      ? pubspec.environment['sdk']
+      : workspaceSdkConstraint;
+
   var subResult = true;
   if (pubspec.isWorkspaceRoot) {
     logger.fine('In a workspace. Recursing through sub-packages...');
     for (final package in pubspec.workspace ?? []) {
-      subResult &= await checkPackage(root: '$root/$package');
+      subResult &= await checkPackage(
+        root: '$root/$package',
+        workspaceSdkConstraint: rootSdkConstraint,
+      );
       logger.info('');
     }
   }
 
   logger.info('Validating dependencies for ${pubspec.name}...');
 
-  final featureSet = featureSetForSdkConstraint(pubspec.environment['sdk']);
+  final sdkConstraint = pubspec.environment['sdk'] ??
+      (pubspec.isInWorkspace ? workspaceSdkConstraint : null);
+  final featureSet = featureSetForSdkConstraint(sdkConstraint);
 
   if (!config.allowPins) {
     checkPubspecForPins(pubspec, ignoredPackages: ignoredPackages);
--- a/lib/src/dependency_validator.dart
+++ b/lib/src/dependency_validator.dart
@@ -90,6 +90,8 @@ Future<bool> checkPackage({required String root}) async {
 
   logger.info('Validating dependencies for ${pubspec.name}...');
 
+  final featureSet = featureSetForSdkConstraint(pubspec.environment['sdk']);
+
   if (!config.allowPins) {
     checkPubspecForPins(pubspec, ignoredPackages: ignoredPackages);
   }
@@ -135,7 +137,9 @@ Future<bool> checkPackage({required String root}) async {
   // export directive.
   final packagesUsedInPublicFiles = <String>{};
   for (final file in publicDartFiles) {
-    packagesUsedInPublicFiles.addAll(getDartDirectivePackageNames(file));
+    packagesUsedInPublicFiles.addAll(
+      getDartDirectivePackageNames(file, featureSet: featureSet),
+    );
   }
   for (final file in publicScssFiles) {
     final matches = importScssPackageRegex.allMatches(file.readAsStringSync());
@@ -201,7 +205,9 @@ Future<bool> checkPackage({required String root}) async {
     if (optionsIncludePackage != null) optionsIncludePackage,
   };
   for (final file in nonPublicDartFiles) {
-    packagesUsedOutsidePublicDirs.addAll(getDartDirectivePackageNames(file));
+    packagesUsedOutsidePublicDirs.addAll(
+      getDartDirectivePackageNames(file, featureSet: featureSet),
+    );
   }
   for (final file in nonPublicScssFiles) {
     final matches = importScssPackageRegex.allMatches(file.readAsStringSync());
--- a/lib/src/import_export_ast_visitor.dart
+++ b/lib/src/import_export_ast_visitor.dart
@@ -13,7 +13,13 @@ import 'package:pub_semver/pub_semver.dart';
 /// Without this, [parseString] falls back to the newest language version the
 /// analyzer knows about, which may be unreleased and reject valid code.
 FeatureSet featureSetForSdkConstraint(VersionConstraint? sdkConstraint) {
-  final min = sdkConstraint is VersionRange ? sdkConstraint.min : null;
+  if (sdkConstraint == null) return FeatureSet.latestLanguageVersion();
+
+  final Version? min = switch (sdkConstraint) {
+    Version version => version,
+    VersionRange range => range.min,
+    _ => null,
+  };
   if (min == null) return FeatureSet.latestLanguageVersion();
 
   return FeatureSet.fromEnableFlags2(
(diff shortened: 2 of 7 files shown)


test/executable_test.dart
```diff
@@ -308,6 +308,30 @@ void main() {
...
+              void log(final Logger logger, {final String message = ''}) {
+                logger.info(message);
+              }
+            '''),
+            ),
+          ]),
+        ],
+      );
+
+      expect(result.stdout, isNot(contains('Error parsing')));
```
engops-wk: This end-to-end test only fails when the analyzer's newest language version happens to reject `final` parameters, so it will silently stop guarding the regression on future analyzer versions. Please add direct unit tests for `featureSetForSdkConstraint` (bounded `^3.6.0`, pinned `3.6.0`, `any`/unbounded, and `null`) asserting the resulting language version / a successful `parseString` of `final`-parameter code.

lib/src/import_export_ast_visitor.dart
```diff
@@ -1,16 +1,37 @@
...
 /// provided dart file
-Set<String> getDartDirectivePackageNames(File file) {
+Set<String> getDartDirectivePackageNames(File file, {FeatureSet? featureSet}) {
   ParseStringResult parsed;
   try {
-    parsed = parseString(content: file.readAsStringSync(), path: file.path);
+    parsed = parseString(
+      content: file.readAsStringSync(),
+      path: file.path,
+      featureSet: featureSet,
```
engops-wk, line 40: Pinning to the declared *minimum* SDK makes parsing stricter than before, so any file using syntax newer than a stale lower bound (generated/fixture files, analysis-excluded files) now hard-`exit(1)`s the whole run instead of parsing. Please add a fallback: catch the `ArgumentError` and retry `parseString` with `FeatureSet.latestLanguageVersion()` before printing and exiting.

lib/src/dependency_validator.dart
```diff
@@ -90,6 +90,8 @@ Future<bool> checkPackage({required String root}) async {
 
   logger.info('Validating dependencies for ${pubspec.name}...');
 
+  final featureSet = featureSetForSdkConstraint(pubspec.environment['sdk']);
```
engops-wk: In a workspace, sub-packages are recursed via `checkPackage` above, but sub-package pubspecs typically omit `environment` (`resolution: workspace` inherits it from the root), so their `featureSet` silently falls back to latest and the fix does not apply. Please resolve the SDK constraint from the workspace root when the package is `isInWorkspace` and has no `environment['sdk']`.

lib/src/import_export_ast_visitor.dart
```diff
@@ -1,16 +1,37 @@
...
 import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:pub_semver/pub_semver.dart';
+
+/// Builds the feature set matching the language version a package declares,
+/// the same way the analyzer derives it from the SDK constraint.
+///
+/// Without this, [parseString] falls back to the newest language version the
+/// analyzer knows about, which may be unreleased and reject valid code.
+FeatureSet featureSetForSdkConstraint(VersionConstraint? sdkConstraint) {
+  final min = sdkConstraint is VersionRange ? sdkConstraint.min : null;
```
engops-wk: `sdkConstraint is VersionRange` misses the pinned-`Version` case (`sdk: 3.6.0`), since `Version` is not a `VersionRange` — those packages silently fall back to the newest language version, which is the bug this PR fixes. Handle `Version` explicitly (or check `is Version` first) and consider truncating pre-release minimums (e.g. `3.7.0-dev` should map to 3.7, matching analyzer behavior).

engops-wk: Correction on my own comment: in `pub_semver`, `Version implements VersionRange` (with `min == max == this`), so pinned `sdk: 3.6.0` is already handled by the `is VersionRange` check. Retracting this as non-blocking — the pre-release truncation nit (`3.7.0-dev` → 3.7) is optional polish.

engops-wk (commented):
## Review round 1 — changes requested (posted as COMMENT; GitHub blocks REQUEST_CHANGES on one's own PR)

The diagnosis and the general approach are right: `parseString` with no `featureSet` defaults to the newest (possibly unreleased) language version the analyzer bundles, and deriving the version from the package's own SDK constraint is exactly how the analyzer determines a file's language version. `pub_semver` is already a direct dependency, so the new import is fine, and the CHANGELOG entry is accurate.

Three blocking concerns before merge:

1. **New hard-failure mode (reliability).** Previously the tool parsed permissively with the newest grammar; now it parses with the package's *minimum* SDK. Any file whose syntax is newer than a stale lower bound (generated code, test fixtures, files normally excluded from analysis) will now throw `ArgumentError` and `exit(1)`, taking down the entire validation run. This is a strictness regression for a lint-adjacent tool that most consumers run in CI. Recommend a retry-with-latest fallback in `getDartDirectivePackageNames` so we only get the benefit of the pinned version, never a new failure.

2. **Test does not durably guard the fix.** The new `executable_test.dart` case passes today only because the currently-bundled analyzer's latest language version rejects `final` parameters. Once that changes, the test passes for the wrong reason and the regression can silently return. `featureSetForSdkConstraint` is a pure function — please cover it directly with unit tests (including the `null`/unbounded fallback path documented in the QA notes, which nothing currently exercises).

3. **Workspace sub-packages skip the fix.** `checkPackage` recurses into workspace members, but members with `resolution: workspace` normally omit `environment`, so they fall back to latest and remain broken. Since workspace support is a headline 5.0.0 feature, this should be handled (or explicitly deferred with a comment plus a follow-up ticket).

Non-blocking nits: consider truncating pre-release SDK minimums (`>=3.7.0-dev` → language version 3.7) to match analyzer behavior; a one-line comment noting why `fromEnableFlags2` with empty flags is used (rather than `FeatureSet.fromEnableFlags`) would help future maintainers.

Happy to re-review quickly once the fallback and the unit tests are in.

engops-wk (commented):
## Review round 2 — Approved ✅ (posted as COMMENT; GitHub blocks self-approval)

All four round-1 items are addressed, and the round-2 changes hold up under a fresh read of the full files (not just the diff). **No blocking issues. Ready to merge.**

**Round-1 follow-through**
1. **Fragile end-to-end-only coverage** → resolved. `test/import_export_ast_visitor_test.dart` now unit-tests `featureSetForSdkConstraint` across bounded `^3.6.0`, pinned `3.6.0`, `any`, `null`, and a pre-release (`3.7.0-dev`) minimum, and asserts `final`-parameter code parses under the derived feature set. Coverage no longer depends on the installed analyzer happening to reject `final` params.
2. **Hard `exit(1)` regression risk from pinning to a stale minimum SDK** → resolved. `getDartDirectivePackageNames` retries with `FeatureSet.latestLanguageVersion()` before printing/exiting, with a dedicated test using 3.8-only null-aware element syntax (`[?1]`) under a 3.6-pinned feature set. Definite assignment of `parsed` still holds because `exit` returns `Never`, and the file read was correctly hoisted so content isn't read twice.
3. **Workspace sub-packages silently losing the fix** → resolved. `checkPackage` threads `workspaceSdkConstraint` through recursion, `sdkConstraint` falls back to it when `isInWorkspace`, and `test/workspace_test.dart` covers a sub-package with `environment` omitted (via `omitSubpackageEnvironment` in `test/utils.dart`).
4. **Pinned `Version` constraint** → previously retracted (`Version implements VersionRange` in pub_semver); the `switch` now handles `Version` explicitly anyway, and the pre-release truncation nit was picked up too.

**Dependency hygiene checked:** `pub_semver` is already a direct dependency and `test_descriptor` a dev dependency, so the new imports won't make dependency_validator flag itself.

**Non-blocking follow-ups** (fine to land as-is):
- `lib/src/dependency_validator.dart`: `rootSdkConstraint` and `sdkConstraint` are near-duplicate derivations. Computing `sdkConstraint`/`featureSet` once above the workspace recursion and passing it down would remove the duplication and also let a *nested* workspace root that omits `environment.sdk` inherit from its parent instead of resetting to `null`.
- `lib/src/import_export_ast_visitor.dart`: the fallback recovers silently; a `fine`-level log ("re-parsed <path> at latest language version") would make this debuggable in the field.
- `test/utils.dart`: `environment: {}` followed by `remove('environment')` is indirect; having `PubspecToJson.toJson()` omit an empty `environment` map would express intent more directly.
- `test/import_export_ast_visitor_test.dart`: `Feature.null_aware_elements` is a recent analyzer constant while the pubspec allows `analyzer >=7.1.0`; test-only, but it would break a lower-bound resolution build.
- `test/workspace_test.dart` calls `checkPackage` in-process, so a parse failure would `exit(1)` the whole test runner rather than fail a test. Pre-existing, but worth replacing `exit` in library code with a returned failure result eventually.
- PR description still references a regression test in `test/executable_test.dart` that is no longer in the diff — please update the body before merge.

matthewnitschke-wk:
Duplicated by https://github.com/Workiva/dependency_validator/pull/187

